PR #40, opened 2017-09-28
Add extra-data sources

alexlarsson:
All this does is construct a finish arg, but it makes it a lot nicer to create extra-data using manifest. Additionally, it allows you to create per-arch extra data if you set only-arches on the source.


alexlarsson:
@rh-atomic-bot r+

PR #41, opened 2017-09-28
Add appdata-license field

alexlarsson:
This lets you modify the project_license field in the appdata file.
This is useful because appdata files from upstream generally only
contain license information for the app itself, whereas the
bundled app may contain other code with additional licenses.


alexlarsson:
@rh-atomic-bot r+

PR #43, opened 2017-10-05
Automatically compress debug info unless no-debuginfo-compression is set

alexlarsson:
This uses eu-elfcompress to compress the debuginfo. We use the older
zlib-gnu compression format which is older and has more widespread
support.


alexlarsson:
@rh-atomic-bot r+

PR #44, opened 2017-10-05
Accept .metainfo.xml appdata extensions

alexlarsson:
We find these either in the appdata or the metainfo directory.
However, we always move it to the appdata dir with the appdata
extension, because that means older versions of appstream-compose
will always pick it up.


alexlarsson:
bot, retest

alexlarsson:
bot, retest

alexlarsson:
@rh-atomic-bot r+

Issue #6, opened 2017-08-25
flatpak-builder: allow to shallow clone git repos


alexlarsson:
_From @alexlarsson on May 19, 2016 15:32_

_From @hbons on February 27, 2016 21:39_

Git repos can be huge, and when building you may not be interested in the full history.
Maybe a property in the json file (or even changing the default) to allow shallow cloning would save a lot of time and bandwidth.

_Copied from original issue: alexlarsson/xdg-app#112_

_Copied from original issue: flatpak/flatpak#15_

alexlarsson:
Hmm, I can see how this could be useful. But its also somewhat tricky. For the initial pull things are fine, but on an update the depth is stored, so the new pull will drop all local history other than the new head of the branch you pulled.

If you're only pulling a repo once this is not really a problem. However, with xdg-app you can have a single download cache shared by multiple json files, or the same json file can refer to the same git repo (either directly or via submodules) which could be problematic if those refered to different branches (or direct commit ids).

I wish there was a way to do an initial shallow clone, but then allow further pulls to just get all changes between your current commit and the new HEAD.

alexlarsson:
_From @hbons on February 29, 2016 13:21_

I think recent Git releases fixed this. You can do a shallow clone and then pull whatever else you need later.

alexlarsson:
There was some kind of fix, but I don't think it does quite what we want here anyway.

alexlarsson:
_From @hbons on February 29, 2016 15:35_

Shallow clones are the default setting in SparkleShare and it works great, so it might be worth looking into. :)

alexlarsson:
_From @jcelerier on June 1, 2016 9:47_

+1 for shallow clones

alexlarsson:
_From @albfan on March 11, 2017 10:39_

Any little test that can be configured to implement this? Like a flatpak project with a dependency with more than one commit. Doing a shallow copy from dep and commiting another change to it should be enough to test if this will work. 

I'm trying to install gnome-todo with flatpak and there's a huge repos gnome-online-accounts and evolution-data-server that has no sense to download entirely.

Moreover when I have them downloaded to use with jhbuild. is there too an option to reuse existing repo clones?

Another problem related is on a unsuccesful build (interrunped) flatpak cannot reuse clone_tmp dir and fails (maybe that can be solved too here or another issue is needed)

alexlarsson:
_From @albfan on March 11, 2017 10:57_

Humm, shallow copies fails

[...]

PR #47, opened 2017-10-12
Do shallow fetches when mirroring git repos

alexlarsson:
When downloading a git repo we try to do a shallow (depth=1) fetch of only the specified target. This normally works fine for branches and tags, but if the commit is a raw SHA1 then it fails, because we can only request refs from the remote.
    
We handle this by doing an ls-remote and seeing if the specified target is either a (possilby partial) ref, or a commit id that a remote ref is pointing at. If it is, we pull that ref only. If not, then we fall back to a full fetch.

I believe this is the best we can do to fix   https://github.com/flatpak/flatpak-builder/issues/6

### DIFF
--- a/src/builder-flatpak-utils.c
+++ b/src/builder-flatpak-utils.c
@@ -519,6 +519,7 @@ flatpak_quote_argv (const char *argv[])
 gboolean
 flatpak_spawn (GFile       *dir,
                char       **output,
+               GSubprocessFlags flags,
                GError     **error,
                const gchar *argv0,
                va_list      ap)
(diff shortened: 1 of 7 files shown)


albfan:
I'm thinking of how to test this.

A flatpak app that has some dep, and change to wip branch using a new wip branch for that dep

Would be better to define a test repo to add this to test suite

alexlarsson:
I tried with this:

[org.test.git.json](https://github.com/flatpak/flatpak-builder/files/1386787/org.test.git.json.txt)

alexlarsson:
I've been running this locally for a while, and it seems to work. Lets merge it.
@rh-atomic-bot r+

PR #49, opened 2017-10-19
Change how the cache works

alexlarsson:
Previously we chained all the state via the internal sha256 checksum, extracting periodic checksums. We now reset the checksum for each state and instead feed it the previous stage checksum to chain things.
    
This means we have different cache checksum values, but the same chaining behaviour. The advantage being that the checksum is re-startable from any point, if we know the parent checksum.

This isn't necessary a huge change right now, but it could potentially allow some nice cache features down the line.


alexlarsson:
@rh-atomic-bot r+

PR #50, opened 2017-10-19
Fix checkout of shallow repo.

alexlarsson:
Some manifest (cx.ring.RingGnome) had a git commit id that
was only tip of a non-branch ref (refs/changes/51/8051/8), which
was not cloned by a regular clone, so we need to switch to
a clone --mirror.


alexlarsson:
@rh-atomic-bot r+

PR #51, opened 2017-10-26
run: Don't pass --die-with-parent when we --run

alexlarsson:
In this case we will exec rather than fork + exec, so we don't
want to use prctl, as that can cause the app to die if any
*thread* in the parent dies (rather then the whole process).

In particular, this caused issues for gnome-builder starting
a newly built flatpak:ed app.


chergert:
This appears to fix the issue for me

alexlarsson:
@rh-atomic-bot r+

PR #52, opened 2017-10-27
Work around issues with older versions of git

alexlarsson:
It turns out older versions of git cannot properly check out a commit
if the ref that points to it was not a normal one (branch or tag).
So, we work around this case by detecting it add adding a fake
tag. Also, we change the fake ref we use for commit-only references
to be a regular branch and not a special one for the same reason.

This fixes https://github.com/flatpak/flatpak/issues/1133


alexlarsson:
@rh-atomic-bot r+

PR #55, opened 2017-10-30
More shallow git fixes

alexlarsson:
This adds a way to disable shallow clones, requested by @directhex, also it disables shallow clones for pre-existing pulls that are not shallow, because this broke for the old version of git we have in the build system. Also, when converting to non-shallow we --unshallow in order to actually get all commits.


alexlarsson:
@rh-atomic-bot r+

PR #60, opened 2017-11-05
cache: honor FLATPAK_OSTREE_REPO_MODE when opening cache repo

### LINKED ISSUE #57: Crash with `fgetxattr: No data available` when reusing .flatpak-builder/ from a previous run
In order to not rebuild my dependencies all the time, I let CI cache the `.flatpak-builder` directory. This results in the following crash though:

```
Starting build of org.wxformbuilder.wxFormBuilder
Cache hit for wxWidgets, skipping build
]2;flatpak-builder: Updating wxWidgetsCache miss, checking out last cache hit

(flatpak-builder:266): flatpak-builder-ERROR **: Failed to check out cache: fgetxattr: No data available
./flatpak.sh: line 8:   266 Trace/breakpoint trap   (core dumped) flatpak-builder --repo=repo ./build org.wxformbuilder.wxFormBuilder.json
```

See https://travis-ci.org/wxFormBuilder/wxFormBuilder/jobs/296809091

Would be great if flatpak-builder could handle that situation.

vrutkovs:
Fixes #57

Signed-off-by: Vadim Rutkovsky <vrutkovs@redhat.com>

### DIFF
--- a/src/builder-cache.c
+++ b/src/builder-cache.c
@@ -245,10 +245,16 @@ builder_cache_open (BuilderCache *self,
     {
       g_autoptr(GFile) parent = g_file_get_parent (builder_context_get_cache_dir (self->context));
 
+      OstreeRepoMode mode = OSTREE_REPO_MODE_BARE_USER;
+      const char *mode_env = g_getenv ("FLATPAK_OSTREE_REPO_MODE");
+
+      if (g_strcmp0 (mode_env, "user-only") == 0)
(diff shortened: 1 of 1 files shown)


alexlarsson:
This is not a good idea, the repo mode used for the cache has to match how it is used during e.g. checkout, and some modes (bare-user-only in particular) are unsafe and will risk silently generating broken builds.

PR #62, opened 2017-11-06
Even more fixes, and some cleanups, for git < 1.9.0

alexlarsson:
With this i managed to build my git testcase both with and without --bundle-sources.


alexlarsson:
Testcase: 
[org.test.git.json.txt](https://github.com/flatpak/flatpak-builder/files/1446120/org.test.git.json.txt)

alexlarsson:
@rh-atomic-bot r+

PR #64, opened 2017-11-13
Add append-pkg-config option

alexlarsson:
This also fixes the checksum for append-path and append-ld-library-path


alexlarsson:
@rh-atomic-bot r+

PR #65, opened 2017-11-15
Add support for running tests

alexlarsson:
Modules that say "run-tests": true, will run tests after installation,
unless disabled by --disable-tests.

The tests run by default are make check or ninja test, however you
can control the make/ninja target with test-rule, or supply a list
of commands with test-commands. There is also a test-args argument
in build-options, which you can use to give e.g. network access.

The tests are run with readonly access to the install directory, so
they cannot affect the build results.

This depends on https://github.com/flatpak/flatpak/pull/1172 for the --readonly flag.

### DIFF
--- a/doc/flatpak-builder.xml
+++ b/doc/flatpak-builder.xml
@@ -204,6 +204,14 @@
             </varlistentry>
 
             <varlistentry>
+                <term><option>--disable-tests</option></term>
+
+                <listitem><para>
+                    Don't run any of the tests.
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
                 <term><option>--run</option></term>
 
                 <listitem><para>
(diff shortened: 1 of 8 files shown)


alexlarsson:
@chergert opinions?

chergert:
I think my only concern is that I might need to start sniffing the flatpak-builder version from Builder to --disable-tests when building dependencies. But that isn't that big of a deal and we should probably start doing that anyway.

I wonder if we should consider something like `flatpak-builder --capabilities` to return a JSON object of what features the `flatpak-builder` supports.

chergert:
Okay, I've tested this with Builder's flatpak-builder json and it seems to work well for our dependencies that don't require access to a display. So it seems useful to land this, and then perhaps figure out the display stuff next.

alexlarsson:
@rh-atomic-bot r+

PR #67, opened 2017-11-15
Add --state-dir option

alexlarsson:
This allows using a shared state dir for downloads and build caches.


alexlarsson:
@rh-atomic-bot r+

PR #68, opened 2017-11-24
Add --enable-p2p configure option

pwithnall:
This got omitted when flatpak-builder was split from flatpak.git. Also fix clean builddir ≠ srcdir builds.


alexlarsson:
@rh-atomic-bot r+

PR #75, opened 2017-12-07
Add more checksum types for files and archives

alexlarsson:
Many upstreams don't use sha256, some use even stronger checksums like
sha512, and its nice to be able to use these. Some system uses
weaker checksums, which you can work around by recomputing your own,
but sometimes that is a bit painful, for example when you're
auto-generating flatpak-builder manifests based on some other format
such as npm lock files.

This adds all the checksum types that GChecksum supports in the
glib version we currently use: md5, sha1, sha256, sha512


src/builder-flatpak-utils.c
```diff
@@ -67,10 +67,13 @@ flatpak_write_update_checksum (GOutputStream  *out,
                                gconstpointer   data,
                                gsize           len,
                                gsize          *out_bytes_written,
-                               GChecksum      *checksum,
+                               GChecksum     **checksums,
+                               int             n_checksums,
```
pwithnall: `gsize` for array lengths, not `int`.

src/builder-context.c
```diff
@@ -368,7 +369,7 @@ builder_context_download_uri (BuilderContext *self,
   if (self->sources_urls != NULL)
     {
       g_autofree char *base_name = g_path_get_basename (soup_uri_get_path (original_uri));
-      g_autofree char *rel = g_build_filename ("downloads", sha256, base_name, NULL);
+      g_autofree char *rel = g_build_filename ("downloads", checksums[0], base_name, NULL);
```
pwithnall: What’s special about index 0 here?

src/builder-flatpak-utils.c
```diff
@@ -67,10 +67,13 @@ flatpak_write_update_checksum (GOutputStream  *out,
...
                                GCancellable   *cancellable,
                                GError        **error)
 {
+  int i;
```
pwithnall: `gsize`

src/builder-flatpak-utils.c
```diff
@@ -82,16 +85,17 @@ flatpak_write_update_checksum (GOutputStream  *out,
...
 
   return TRUE;
 }
 
 gboolean
 flatpak_splice_update_checksum (GOutputStream  *out,
                                 GInputStream   *in,
-                                GChecksum      *checksum,
+                                GChecksum     **checksums,
+                                int             n_checksums,
```
pwithnall: `gsize`

src/builder-flatpak-utils.h
```diff
@@ -94,14 +94,16 @@ gboolean flatpak_write_update_checksum (GOutputStream  *out,
                                         gconstpointer   data,
                                         gsize           len,
                                         gsize          *out_bytes_written,
-                                        GChecksum      *checksum,
+                                        GChecksum     **checksums,
+                                        int             n_checksums,
```
pwithnall: `gsize`

src/builder-flatpak-utils.h
```diff
@@ -94,14 +94,16 @@ gboolean flatpak_write_update_checksum (GOutputStream  *out,
...
                                         GCancellable   *cancellable,
                                         GError        **error);
 
 
 gboolean flatpak_splice_update_checksum (GOutputStream  *out,
                                          GInputStream   *in,
-                                         GChecksum      *checksum,
+                                         GChecksum     **checksums,
+                                         int             n_checksums,
```
pwithnall: `gsize`

src/builder-source-archive.c
```diff
@@ -677,13 +738,34 @@ builder_source_archive_class_init (BuilderSourceArchiveClass *klass)
                                                         "",
                                                         NULL,
                                                         G_PARAM_READWRITE));
+  g_object_class_install_property (object_class,
```
pwithnall: I realise the existing `sha256` property doesn’t have any documentation, but it would be nice to add at least a comment saying what the checksum is a checksum of, for each of these.

src/builder-source-file.c
```diff
@@ -534,13 +595,34 @@ builder_source_file_class_init (BuilderSourceFileClass *klass)
                                                         "",
                                                         NULL,
                                                         G_PARAM_READWRITE));
+  g_object_class_install_property (object_class,
```
pwithnall: Same here with the documentation.

src/builder-utils.c
```diff
@@ -1615,6 +1615,95 @@ builder_maybe_host_spawnv (GFile                *dir,
...
 
+int
+builder_get_all_checksums (const char *checksums[BUILDER_CHECKSUMS_LEN],
+                           GChecksumType checksums_type[BUILDER_CHECKSUMS_LEN],
+                           const char *md5,
+                           const char *sha1,
+                           const char *sha256,
+                           const char *sha512)
+{
+  int i = 0;
```
pwithnall: `gsize`

src/builder-utils.c
```diff
@@ -1615,6 +1615,95 @@ builder_maybe_host_spawnv (GFile                *dir,
...
+
+  /* The order here is such that the first one is the
+     one to use by default if using a single one. That is
+     the longest checksum except it defaults to sha256
+     which was historically the one flatpak-builder used */
```
pwithnall: ⇒ gtk-doc comment, ideally.

src/builder-utils.c
```diff
@@ -1615,6 +1615,95 @@ builder_maybe_host_spawnv (GFile                *dir,
   return flatpak_spawnv (dir, output, 0, error, argv);
 }
 
+int
```
pwithnall: `gsize`

src/builder-utils.c
```diff
@@ -1615,6 +1615,95 @@ builder_maybe_host_spawnv (GFile                *dir,
...
+      checksums_type[i] = G_CHECKSUM_SHA1;
+      i++;
+    }
+
+  if (md5 != NULL && *md5 != 0)
+    {
+      checksums[i] = md5;
+      checksums_type[i] = G_CHECKSUM_MD5;
+      i++;
+    }
```
pwithnall: Do you want to add SHA-384 support too? It got added in GLib 2.52.

pwithnall: (Does anyone use it in the wild?)

src/builder-utils.c
```diff
@@ -1615,6 +1615,95 @@ builder_maybe_host_spawnv (GFile                *dir,
...
+}
+
+static gboolean
+compare_checksum (const char *name,
+                  const char *expected_checksum,
+                  GChecksumType checksum_type,
+                  const char *measured_checksum,
+                  GError **error)
+{
+  const char *type_names[] = { "md5", "sha1", "sha256", "sha512", "sha384" }; /* In GChecksumType order */
```
pwithnall: I’d add a precondition check that `checksum_type` doesn’t index off the end of `type_names` here, just in case new checksum types get added to GLib before this function is updated.

src/builder-utils.c
```diff
@@ -1615,6 +1615,95 @@ builder_maybe_host_spawnv (GFile                *dir,
...
+
+gboolean
+builder_verify_checksums (const char *name,
+                          char *data,
+                          gsize len,
+                          const char *checksums[BUILDER_CHECKSUMS_LEN],
+                          GChecksumType checksums_type[BUILDER_CHECKSUMS_LEN],
+                          GError **error)
+{
+  int i;
```
pwithnall: `gsize`

src/builder-utils.c
```diff
@@ -1615,6 +1615,95 @@ builder_maybe_host_spawnv (GFile                *dir,
...
+                   expected_checksum, measured_checksum);
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+gboolean
+builder_verify_checksums (const char *name,
+                          char *data,
```
pwithnall: `const`?

src/builder-utils.c
```diff
@@ -1615,6 +1615,95 @@ builder_maybe_host_spawnv (GFile                *dir,
...
+                          gsize len,
+                          const char *checksums[BUILDER_CHECKSUMS_LEN],
+                          GChecksumType checksums_type[BUILDER_CHECKSUMS_LEN],
```
pwithnall: `const`

src/builder-utils.c
```diff
@@ -1653,10 +1743,15 @@ builder_download_uri (SoupURI        *uri,
   g_autoptr(GFileOutputStream) out = NULL;
   g_autoptr(GFile) tmp = NULL;
   g_autoptr(GFile) dir = NULL;
-  g_autoptr(GChecksum) checksum = g_checksum_new (G_CHECKSUM_SHA256);
+  g_autoptr(GPtrArray) checksum_array = g_ptr_array_new_with_free_func ((GDestroyNotify)g_checksum_free);
   DownloadPromptData progress_data = {0};
   g_autofree char *basename = g_file_get_basename (dest);
   g_autofree char *template = g_strconcat (".", basename, "XXXXXX", NULL);
+  int i;
```
pwithnall: `gsize`

src/builder-utils.c
```diff
@@ -1615,6 +1615,95 @@ builder_maybe_host_spawnv (GFile                *dir,
...
+  /* The order here is such that the first one is the
+     one to use by default if using a single one. That is
+     the longest checksum except it defaults to sha256
+     which was historically the one flatpak-builder used */
+
+  if (sha256 != NULL && *sha256 != 0)
+    {
+      checksums[i] = sha256;
+      checksums_type[i] = G_CHECKSUM_SHA256;
+      i++;
```
pwithnall: It would be good to tie the definition of `BUILDER_CHECKSUMS_LEN` to this function, so that they can’t be adjusted out of step.

`g_assert (i < BUILDER_CHECKSUMS_LEN)`?

doc/flatpak-manifest.xml
```diff
@@ -531,10 +531,22 @@
                         <term><option>url</option> (string)</term>
                         <listitem><para>The URL of a remote archive that will be downloaded. This overrides path if both are specified.</para></listitem>
                     </varlistentry>
+                    <varlistentry>
+                        <term><option>md5</option> (string)</term>
+                        <listitem><para>The md5 checksum of the file, verified after download</para></listitem>
+                    </varlistentry>
```
pwithnall: It would be good to mention somewhere that SHA-256 is still the preferred/default checksum type, and that MD5 and SHA-1 are discouraged and shouldn’t be relied on if possible.

src/builder-context.c
alexlarsson: The list is ordered such that the first is the "default" to use for when you only can use one checksum, such as in the filename in the download cache. See the comments in builder_get_all_checksums.

src/builder-flatpak-utils.c
alexlarsson: Well, its at most 4, but i guess it looks nicer...

src/builder-source-archive.c
alexlarsson: These are all internal properties just so that the json-glib serialization stuff works. I don't think documenting them here is useful, the useful place is the user facing documentation.

src/builder-utils.c
alexlarsson: I initially added it, but removed it because it needs a newer glib than what we require (and is available in rhel7/xenial).

src/builder-utils.c
alexlarsson: How can an enum be const?

src/builder-flatpak-utils.c
pwithnall: It also won’t give you problems with signed vs unsigned. I think it’s just a good habit to get into, rather than having to think about the maximum possible length of each array you come across.

src/builder-utils.c
pwithnall: Ah, that makes sense.

src/builder-utils.c
pwithnall: `const` array

pwithnall:
@rh-atomic-bot r+ d8f46b5c2f0c2f6f5cf2b8ef809415307b865230

pwithnall:
sad panda 🐼

alexlarsson:
Added you as collaborator.

@rh-atomic-bot r+

alexlarsson:
CI seems busted, lets manually merge this for now.

PR #77, opened 2017-12-07
Support including json snippet for sources

alexlarsson:
Support including json snippet for sources

This allows inclusion of sources from an external json source similar to how we do it for modules.
    
For example, you can use:

    "sources": [
        {
            "type": "shell",
            "commands": [ "echo BEFORE include" ]
        },
        "include.json",
        {
            "type": "shell",
            "commands": [ "echo AFTER include" ]
        }
    ]
    
with include.json containing:

    [
        {
            "type": "shell",
            "commands": [ "echo Shell 1" ]
        },
        {
            "type": "shell",
            "commands": [ "echo Shell 2" ]
        }
    ]
    
This is very useful in the case where the included file is  auto-generated from some other source, such as an npm lockfile.

### DIFF
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -376,8 +376,12 @@
                     <listitem><para>If true, skip this module</para></listitem>
                 </varlistentry>
                 <varlistentry>
-                    <term><option>sources</option> (array of objects)</term>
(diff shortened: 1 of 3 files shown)


alexlarsson:
Been testing this all day with npm stuff, and its working great.
@rh-atomic-bot r+

PR #78, opened 2017-12-08
Fix file sources that have no sha256 checksums (but have others)

alexlarsson:
This was missing from the previous commit, and makes file sources
work just like archive sources.


pwithnall:
@rh-atomic-bot r+ 9c350402600df526c98d25b7ede212a36ed8289c

PR #79, opened 2017-12-08
Don't require checksum for inline data

alexlarsson:
This fixes a regression from ac38819106eb039252f0ae45f709896beac70545


pwithnall:
@rh-atomic-bot r+ 004c1cc9834ab8213ee5f8d85982aa4069f2b6a1

PR #80, opened 2017-12-11
Use canonical formats in cache and use bare-user-only repo

alexlarsson:
This means builds work on filesystems without xattrs, and that the cache repo modes matches the outpuf from flatpak build-export, which can be useful in some cases.


alexlarsson:
@rh-atomic-bot r+

PR #83, opened 2017-12-13
appdata: Only rename complete matches of the id

alexlarsson:
We don't want to rename partial hits of the desktop id, nor
inside any other tag.

Fixes https://github.com/flatpak/flatpak/issues/1118


alexlarsson:
@rh-atomic-bot r+

PR #86, opened 2017-12-18
Fix some NULL references in --install without --repo

alexlarsson:


src/builder-main.c
```diff
@@ -850,9 +850,10 @@ main (int    argc,
       g_auto(GStrv) exclude_dirs = builder_manifest_get_exclude_dirs (manifest);
       GList *l;
 
-      export_repo = g_file_new_for_path (opt_repo);
+      if (opt_repo)
+        export_repo = g_file_new_for_path (opt_repo);
       if (opt_install && export_repo == NULL)
```
pwithnall: Would be less confusing if this second `if` were an `else if (opt_install)`.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #94, opened 2018-01-08
docs: fix "concatinated" and "inbetween" typos

wjt:
"In between" is two separate words, but I think rewording it is clearer.


pwithnall:
@rh-atomic-bot r+ 345c5a50fa6063002db495fe69f8f6767543cce2

PR #97, opened 2018-01-10
context: check state-dir when locating sources

chergert:
If --state-dir is provided, we might want to find the sources directory
within that path. This fixes an issue with Builder where it fails to
locate the source-dir and therefore always pings the remote git URL.

### DIFF
--- a/src/builder-context.c
+++ b/src/builder-context.c
@@ -302,6 +302,19 @@ builder_context_find_in_sources_dirs_va (BuilderContext *self,
 {
   int i;
 
+  if (self->state_dir != NULL)
+    {
+      g_autoptr(GFile) local_file = NULL;
+      va_list args2;
+
+      va_copy (args2, args);
+      local_file = flatpak_build_file_va (self->state_dir, args2);
+      va_end (args2);
+
+      if (g_file_query_exists (local_file, NULL))
+        return g_steal_pointer (&local_file);
+    }
+
   if (self->sources_dirs == NULL)
     return NULL;
 


alexlarsson:
I don't understand this. If you specify --state-dir then that is used instead of `$cwd/.flatpak-builder` for storing downloads. For instance git downloads in `$statedir/git/$mangled_uri. Additionally we can supply --extra-sources, which can point to a pre-supplied directory of downloaded sources/repos. Such directories are pure read only, and if anything is there we use it as-is, but if not we continue to the state-dir.

But this patch treats the state dir as an --extra-sources, which means we will always use it read-only, so we will e.g. never ever update a git repo once pulled. This would make e.g. the nightly gnome apps build never rebuild newer versions.

chergert:
Hrmm, yeah that might explain some things I need to troubleshoot still.

The problem I'm trying to fix in Builder is that we need to be able to say "only ever communicate with the repository peer if you have nothing locally". Otherwise, Builder is communicating with peers a lot and it really slows down the process of getting into your project and hacking.

I think it *used* to work, but we use `--disable-updates` and `--download-only`. We provide `--state-dir` if we have a new enough flatpak-builder.

alexlarsson:
--disable-updates should do that, it means "clone if you have nothing, but don't pull if you have a repo already"

chergert:
It's broken then.

chergert:
builder-git.c:519 will have `update=FALSE` and `cached_git_dir=NULL` in this case.

alexlarsson:
Closed in favour of #98

PR #98, opened 2018-01-11
Fix git downloads with --disable-updates

alexlarsson:
We regressed in --disable-updates in for git sources, as they were always downloaded, as per https://github.com/flatpak/flatpak-builder/pull/97

This fixes that, and also makes intial download a bit more robust by using a temporary destination.


chergert:
Seems to work for me, and code looks good.

alexlarsson:
@rh-atomic-bot r+

PR #100, opened 2018-01-12
Fix running flatpak-builder from inside flatpak

chergert:
This series of patches allows us to be able to run a bundled `flatpak-builder` with Builder and execute the necessary `flatpak` operations on the host.

There may still be some operations that are direct subprocesses (build-export, etc), but I couldn't justify why they should run on the host so I left them as is.

One of the interesting fixes in this series is the GUnixFDList fix. Without that, running on the host would indefinitely block while waiting for an async splice to occur.

### DIFF
--- a/src/builder-main.c
+++ b/src/builder-main.c
@@ -830,6 +830,7 @@ main (int    argc,
         {
           if (!builder_maybe_host_spawnv (NULL,
                                           NULL,
+                                          0,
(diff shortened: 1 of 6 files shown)


alexlarsson:
bot, add author to whitelist

alexlarsson:
@rh-atomic-bot r+

PR #107, opened 2018-02-16
Add support for -y, passed on to all install commands

alexlarsson:
This allows you to run --install-from-deps=foo -y to always install all
dependencies. Useful for auto-builders like flathub.


mwleeds:
@rh-atomic-bot r+

PR #118, opened 2018-03-27
ccache: Enforce disabled if not enabled

alexlarsson:
This makes sure that ccache is disabled if you don't enable
it on the commandline. This makes sense, because we don't
have any persistant location for the ccache files anyway.

Additionally this is done to work around a race condition initializing
~/.ccache/ccache.conf that causes meson builds to error out.


alexlarsson:
@rh-atomic-bot

alexlarsson:
@rh-atomic-bot r+

PR #119, opened 2018-03-27
build options: Add prepend-*-path options

alexlarsson:
This is useful for example if you're using a sdk-extension that
is supposed to override some binaries from /usr/bin.


alexlarsson:
@rh-atomic-bot r+

PR #120, opened 2018-03-27
Add --no-shallow-clone for git mirroring

alexlarsson:
This is useful e.g. when doing mirroring only, or if your git
has issues with shallow clones.


alexlarsson:
@rh-atomic-bot r+

PR #121, opened 2018-03-27
Add inherit-sdk-extensions property

alexlarsson:
This is similar to inherit-extensions, but the extensions
are not also inherited into the platform when it is created.


alexlarsson:
@rh-atomic-bot r+

PR #122, opened 2018-03-27
Don't warn for // and __ key prefixes

alexlarsson:
This means you can make valid json comments like so:

 "//": "Some comment",


alexlarsson:
@rh-atomic-bot r+

PR #124, opened 2018-04-06
Fix --install to not pass --subpath= to flatpak install

alexlarsson:
This isn't really right, as install doesn't handle an empty subpath
like that. In fact, doing so will break exports.


alexlarsson:
@rh-atomic-bot r+

PR #126, opened 2018-04-09
manifest: Add support for "extension-tag"

smspillaz:
This passes an --extension-tag to flatpak build-init which will
set the "tag" option on the ExtensionOf section in the metadata.

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -62,6 +62,7 @@ struct BuilderManifest
   char           *id_platform;
   char           *branch;
   char           *collection_id;
+  char           *extension_tag;
   char           *type;
   char           *runtime;
   char           *runtime_commit;
@@ -161,6 +162,7 @@ enum {
   PROP_DESKTOP_FILE_NAME_SUFFIX,
   PROP_COLLECTION_ID,
   PROP_ADD_EXTENSIONS,
+  PROP_EXTENSION_TAG,
   LAST_PROP
 };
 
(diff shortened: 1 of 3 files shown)


src/builder-manifest.c
```diff
@@ -1601,6 +1630,9 @@ builder_manifest_checksum (BuilderManifest *self,
   builder_cache_checksum_str (cache, self->base_commit);
   builder_cache_checksum_strv (cache, self->base_extensions);
 
+  if (self->extension_tag != NULL)
+    builder_cache_checksum_str (cache, self->extension_tag);
```
alexlarsson: please use builder_cache_checksum_compat_str() here instead of the if.

src/builder-manifest.c
```diff
@@ -2844,18 +2887,23 @@ builder_manifest_finish (BuilderManifest *self,
...
           extension_metadata_name = g_strdup_printf ("metadata.%s", builder_extension_get_name (e));
           metadata_extension_file = g_file_get_child (app_dir, extension_metadata_name);
           metadata_contents = g_strdup_printf ("[Runtime]\n"
                                                "name=%s\n"
                                                "\n"
                                                "[ExtensionOf]\n"
-                                               "ref=%s\n",
-                                               builder_extension_get_name (e), ref);
+                                               "ref=%s\n"
+                                               "%s\n",
```
alexlarsson: Can you add the \n in maybe_format_extension_tag so we don't leave an unnecessary \n here otherwise.

smspillaz:
Updated

alexlarsson:
@rh-atomic-bot r+

PR #127, opened 2018-04-14
Add YAML support as an alternative to JSON

### LINKED ISSUE #2: Alternative to JSON manifest for flatpak-build?
_From @mdeguzis on July 31, 2017 0:46_

I don't like so much that you can't really avoid having to manually compile dependencies as part of your build. For something like Kodi, that is a huge amount of "modules" you'd have to write in and get all correct vs. just adding a line to debian/control for an existing ffmpeg package.

JSON gets messy, and keeping which bracket/brace goes to which gets out of hand with long files. Contrast this to A Makefile or debian/rules, where things are structure with labels, and allow scenarios where "this depends on that", just as you have modules here. I see the JSON manifest as a barrier for acceptance, I know it is for me, even if I write JSON payloads frequently for my day job as a devop for Hadoop.

Having packaged for Debian/Fedora/Arch Linux for several years, it just seems there is a better way. Any plans for an alternative? JSON generator?

_Copied from original issue: flatpak/flatpak#941_

kirbyfan64:
Closes #2.

Some notes:

- Rather than rebuilding the entire deserialization logic, I simply converted the YAML tree to `JsonNode`s. The advantage is this is far less intrusive, and it piggyback's on json-glib's already-present, battle-tested deserialization logic.
- libyaml is present all the way back to CentOS/RHEL 6 and Ubuntu 14.04. (On the two RH distros, libyaml-devel is in the optional repos, though it seems most \*-devel files are in there as well.)
- That being said, since there was still concern in #2 about portability, I put it behind an autoconf switch. By default, it'll automatically check for libyaml, and enable it if found; it can be forcibly enabled via `--with-yaml` or disabled via `--without-yaml`.

You can just look at the new tests to see the big difference in readability and conciseness in YAML vs JSON, and all for a 124KB shared library. :wink:

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -10,7 +10,7 @@ pkg_install sudo which attr fuse \
     libubsan libasan libtsan elfutils-libelf-devel libdwarf-devel \
     elfutils git gettext-devel libappstream-glib-devel \
     /usr/bin/{update-mime-database,update-desktop-database,gtk-update-icon-cache}
-pkg_install_testing ostree-devel ostree
+pkg_install_testing ostree-devel ostree libyaml-devel
 pkg_install_if_os fedora gjs parallel clang
 pkg_install_builddeps flatpak
 
--- a/configure.ac
+++ b/configure.ac
@@ -146,6 +146,17 @@ AS_IF([test "x$with_dwarf_header" = "xyes"],
        AS_IF([test "x$ac_cv_header_dwarf_h" != "xyes"],
              [AC_MSG_ERROR([dwarf.h is required but was not found])])])
 
+AC_ARG_WITH([yaml],
+  [AS_HELP_STRING([--without-yaml],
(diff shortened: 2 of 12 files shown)


alexlarsson:
bot, add author to whitelist

alexlarsson:
bot, test pull request

alexlarsson:
bot, test pull request

alexlarsson:
@rh-atomic-bot r+

mdeguzis:
Thanks for the PR merge! YAML is much more suited for things like this (configuration) vs. a data-interchange driven serialization format like JSON.

mdeguzis:
Is a generic JSON to YAMl conversion ok, or will I need to parse a few things?

Example:

https://github.com/mdeguzis/LibreGeek-Packaging/blob/flatpak/utils/json-to-yaml.py

PR #128, opened 2018-04-19
Record the built extensions in the metadata

alexlarsson:
This adds a section to the main metadata like:
```
[Build]
built-extension=org.the.App.Locale;
```

This can be used to figure out what refs where built that are
related to the app, for example if you want to bundle them all.


alexlarsson:
@rh-atomic-bot r+

PR #129, opened 2018-04-20
Support using app-extensions during the build

alexlarsson:
This adds add-build-extensions which is similar to add-extensions
except the extension is added at build-init time, so can be
used during the build. It can also optionally be removed after
the build is done.

This depends on the flatpak work in:
  https://github.com/flatpak/flatpak/pull/1598

With this I was able to build the following app which runs 32bit binaries
in a 64bit build:

```json
{
    "app-id": "org.example.Multilib",
    "runtime": "org.freedesktop.Platform",
    "sdk": "org.freedesktop.Sdk",
    "runtime-version": "1.6",
    "command": "/usr/bin/true",
    "add-build-extensions": {
        "org.freedesktop.Platform.Compat32": {
            "directory": "lib/32bit",
            "add-ld-path": "lib",
            "version": "1.6"
        }
    },
    "modules": [
        {
            "name": "test 32bit",
            "buildsystem": "simple",
            "build-commands": [
                "ln -s /app/lib/32bit/lib/ld-linux.so.2 /app/lib/ld-linux.so.2",
                "/app/lib/32bit/bin/echo echoing from 32bit world"
            ]
        }
    ]
}
```

### DIFF
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -193,6 +193,10 @@
                     <listitem><para>This is a dictionary of extension objects. The key is the name of the extension. See below for details.</para></listitem>
                 </varlistentry>
                 <varlistentry>
+                    <term><option>add-build-extensions</option> (objects)</term>
+                    <listitem><para>This is a dictionary of extension objects similar to add-extensions. The main difference is that the extensions are added early and are available for use during the build.</para></listitem>
+                </varlistentry>
+                <varlistentry>
                     <term><option>cleanup</option> (array of strings)</term>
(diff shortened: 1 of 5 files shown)


alexlarsson:
@johnramsden I think this is a step on the way for what you want. I think with the better multilib setup in freedesktop 1.8 it would be even nicer.

johnramsden:
Thanks @alexlarsson, this looks perfect.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@johnramsden 1.8 is at https://gitlab.com/freedesktop-sdk/freedesktop-sdk

PR #131, opened 2018-04-23
Always pass --unshare=network when building sandboxed

alexlarsson:
It may be the case that we've e.g. specified a custom metadata file that
has network access, so always disable it explicitly during the builds.


alexlarsson:
@rh-atomic-bot r+

Issue #133, opened 2018-04-24
Downloading files from a non-supported protocol URL (e.g. svn)


magicmyth:
I've made a flatpak-builder manifest for the GtkRadiant (a ID Tech game engine level editor) app here:
https://github.com/magicmyth/io.github.TTimo.GtkRadiant
Its an early first attempt at it so plenty still to do.

Currently its build process downloads game paks via svn (svn gets built by the manifest as well). I know downloading during the actual build process is not recommended but at the moment that's just how this project works but I do intend to change that part. The problem is these game paks are only distributed over svn or git and must be present during build for GtkRadiant to support the game. I would like to move the grabbing of the game paks to a separate source or module part of the phase and the git based ones should be easy to do as there is support for git URLs built in but as flatpak-builder does not seem to do svn I can't currently do that. I could do a shell type source with the _build-arg --share=network_ and run the svn checkout command that way but would that really be any different to adding network access to the apps build process as I am currently doing?

So I was wondering if it is possible to expose other kinds of file retrieval protocols to builder? Is it just recommended that such scenarios use shell or script types with network access to grab such sources? If so how does one go about exposing that in a way the cache system would continue to function? I imagine there will be other scenarios where developers do not publish archives over http.

TingPing:
>  Is it just recommended that such scenarios use shell or script types with network access to grab such sources?

No. The ideal solution is either `flatpak-builder` understands svn or the files get hosted as an archive somewhere else upstream.

magicmyth:
That's what I expected (and agree). It would be useful if we could extend what kind of file retrieval protocols flatpak-builder can use though (git may not always be king a proprietary shops can often use strange VCs). However having support for svn would greatly expand the existing projects that can be easily supported.

alexlarsson:
Adding a minimal svn source implementation should not be all that hard. Lemme give it a go today.

magicmyth:
Thanks much appreciated. Though now I wish I had asked sooner so I would have saved time figuring out so many parts. Nah learnt a lot from that.

So am I correct in understanding if apps themselves used svn they would still need to either build svn into their flatpak or have it as a runtime extension? If the later I may have a go at making a runtime extension for the FDO runtime that apps can use. Or is that what a base app would be used for?

You are a super star for doing this.

alexlarsson:
Yeah, svn is not part of the current sdk or runtime, so you have to include your own copy.

alexlarsson:
[...]

PR #135, opened 2018-04-25
Add SVN source type

alexlarsson:
This adds support for SVN sources, fixing https://github.com/flatpak/flatpak-builder/issues/133

@magicmyth does this look ok to you?

### DIFF
--- a/doc/flatpak-builder.xml
+++ b/doc/flatpak-builder.xml
@@ -199,7 +199,7 @@
 
                 <listitem><para>
                   Download missing sources, but don't update local mirrors of version control repos. This is useful
-                  to rebuild things but without updating git or bzr repositories from the remote repository.
+                  to rebuild things but without updating git, bzr or svn repositories from the remote repository.
                 </para></listitem>
             </varlistentry>
 
@@ -499,7 +499,7 @@
                 <term><option>--extra-sources=SOURCE-DIR</option></term>
 
                 <listitem><para>
-                  When downloading sources (archives, files, git, bzr), look in this
+                  When downloading sources (archives, files, git, bzr, svn), look in this
                   directory for pre-existing copies and use them instead of downloading.
                 </para></listitem>
             </varlistentry>
@@ -508,7 +508,7 @@
                 <term><option>--extra-sources-url=URL</option></term>
 
                 <listitem><para>
-                  When downloading sources (archives, files, git, bzr), look at this url
+                  When downloading sources (archives, files, git, bzr, svn), look at this url
                   for mirrored downloads before downloading from the original url.
                 </para></listitem>
             </varlistentry>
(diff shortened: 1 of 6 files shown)


alexlarsson:
bot, retest this please

alexlarsson:
I actually want to make a flatpak for flatpak-builder, but that won't have the patch anyway. However, you don't need to rebuild flatpak, only flatpak-builder.

alexlarsson:
Oh, and yeah, that is right, but you have to add a url too, and "revision" is optional (defaults to latest).

magicmyth:
Thanks for getting back quick. I totally would have built flatpak instead of flatpak builder and been scratching my head.

I've just done a quick test with the following manifest:
```
{
  "app-id": "lets.test.svn",
  "runtime": "org.freedesktop.Platform",
  "runtime-version": "1.6",
  "sdk": "org.freedesktop.Sdk",
  "command": "not-real",
  "finish-args": [
     "--filesystem=home"
  ],
  "modules": [
    {
      "name": "radiant-paks",
      "buildsystem": "simple",
      "sources": [
        {
          "type": "svn",
          "url": "svn://svn.icculus.org/gtkradiant-gamepacks/Q1Pack/trunk",
          "revision": "16",
          "dest": "gamepacks/Q1Pack"
        },
        {
          "type": "svn",
          "url": "svn://svn.icculus.org/gtkradiant-gamepacks/Q2Pack/branches/1.4",
          "dest": "gamepacks/Q2Pack"
        },
        {
          "type": "svn",
          "url": "svn://svn.icculus.org/gtkradiant-gamepacks/Q3Pack/trunk",
          "dest": "gamepacks/Q3Pack"
        },
        {
          "type": "archive",
          "url": "https://downloads.sourceforge.net/project/liboauth/liboauth-1.0.3.tar.gz",
          "sha256": "0df60157b052f0e774ade8a8bac59d6e8d4b464058cc55f9208d72e41156811f"
        }
      ],
      "build-commands": [
        "pwd",
        "ls",
        "echo 'Check if the packs are there'",
        "ls gamepacks",
        "exit 1"
      ]
    }
  ]
}
```

and it worked perfectly! The liboauth source is just so I could test mixing the sources in the way I intend to with GtkRadiant. I can confirm the checkout caches are working and as you can see I tested with a revision, branch and trunk checkouts. All worked as expected.

No idea how you made that work so darn quick but thanks!

magicmyth:
I meant to say that when trying to build flatpak-builder there was no `configure` script. Running `./autogen.sh` fixed that but its not mentioned on the README.md so maybe its worth adding that tid bit?

alexlarsson:
The configure script is only in the release tarball, if you're building from git you have to use autogen.sh. This is pretty standard, but maybe the readme should mention it.

alexlarsson:
@rh-atomic-bot r+

PR #136, opened 2018-04-25
Add directory source

alexlarsson:
This adds a source type that copies an entire directory, optionally skipping some files. Additionally it also skips the .flatpak-directory dir and the build dir to avoid weird recursion.

Since we can't really checksum an entire directory a dir source is always rebuilt.

@matthiasclasen you were interested in this, right?


matthiasclasen:
Yes, I was. Thanks!

alexlarsson:
@rh-atomic-bot r+

PR #137, opened 2018-04-26
Set FLATPAK_BUILDER_N_JOBS in build env

alexlarsson:
This is set to what flatpak itself would otherwise use for make -jN
and can be used in e.g. buildsystem simple modules.


alexlarsson:
@rh-atomic-bot r+

PR #139, opened 2018-05-01
Provide right --libdir option when system is multiarch

valentindavid:
This change is required to build against freedesktop-sdk 1.8 which uses multiarch library paths.

### DIFF
--- a/src/builder-module.c
+++ b/src/builder-module.c
@@ -1237,6 +1237,46 @@ shell (GFile          *app_dir,
 static const char skip_arg[] = "skip";
 static const char strv_arg[] = "strv";
 
+static gchar*
+get_libdir(GFile          *app_dir,
+           const char     *module_name,
+           BuilderContext *context,
+           GFile          *source_dir,
+           const char     *cwd_subdir,
+           char          **flatpak_opts,
+           const char     *prefix,
+           GError        **error) {
+  g_autofree char *output = NULL;
+  g_autoptr(GFile) cwd_file = NULL;
+  g_autoptr(GPtrArray) args =
+    setup_build_args (app_dir, module_name, context, source_dir, cwd_subdir, flatpak_opts, NULL, &cwd_file);
+
+  g_ptr_array_add (args, g_strdup ("gcc"));
+  g_ptr_array_add (args, g_strdup ("-print-multiarch"));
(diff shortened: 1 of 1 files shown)


alexlarsson:
Can we instead return NULL for libdir in the non-runtime case and avoid passing it as a configure argument? That way we don't change anything in how current builds are done, which could risk breaking some existing manifest.

src/builder-module.c
```diff
@@ -1237,6 +1237,46 @@ shell (GFile          *app_dir,
...
+  g_ptr_array_add (args, NULL);
+
+  if (!builder_maybe_host_spawnv (cwd_file, &output, 0, error, (const char * const *)args->pdata))
+    {
+      g_prefix_error (error, "module %s: ", module_name);
+      return NULL;
+    }
+
+  {
+    char *new_line = strchr (output, '\n');
```
alexlarsson: Can't you just use g_strchomp (output); here?

src/builder-module.c
```diff
@@ -1237,6 +1237,46 @@ shell (GFile          *app_dir,
...
+           char          **flatpak_opts,
+           const char     *prefix,
+           GError        **error) {
+  g_autofree char *output = NULL;
+  g_autoptr(GFile) cwd_file = NULL;
+  g_autoptr(GPtrArray) args =
+    setup_build_args (app_dir, module_name, context, source_dir, cwd_subdir, flatpak_opts, NULL, &cwd_file);
+
+  g_ptr_array_add (args, g_strdup ("gcc"));
+  g_ptr_array_add (args, g_strdup ("-print-multiarch"));
```
alexlarsson: I'm somewhat worried about relying on gcc features in the runtime for this. I wonder if we can instead store this info in the sdk somehow? We've been talking about other things like this, for instance a default CFLAGS for security flags.

alexlarsson:
Hmm, so, could you add a libdir property to the manifest file and a `/etc/flatpak-builder/defaults.json` file to the sdk, with code in f-b to parse this for the default value of libdir, as well as cflags, cppflags, cxxflags, ldflags. Then any app that don't override these will pick up the default.

valentindavid:
I can try that.

valentindavid:
Please see PR #142.

PR #142, opened 2018-05-02
Load libdir and flags from SDK configuration file

valentindavid:
This is a rewrite of #139 using `/etc/flatpak-builder/defaults.json` configuration file from SDK instead of detecting libdir from GCC.

Should we create BuilderSdkConfig as soon as we have checked out `init` and then pass it around? For the moment I read the configuration every time I need.

Debug directory also needs to be moved. This is in order to provide .Debug extension for Compat extensions in the future.


alexlarsson:
What you want is to load the defaults once and put them in BuilderContext *, and then access this
from e.g. builder_options_get_env(), where the default values are used in the call to builder_options_get_cflags() if the flag is not set. This way it is overridable if the default is not working for some reason.

Same with libdir, i.e. make it a BuildOptions property like e.g. cflags, and make the defaults be the default for that.

valentindavid:
I understand the need for disabling options for SDK configuration. But I think the default behavior should be to concatenate all flags. If you want to just add one CPPFLAGS, you do not want to just disable all CPPFLAGS from the SDK.

Erick555:
+1 for concatenation. It can work like this:
`upstream flags` + `runtime flags` + `app flags`

In the other thread I proposed to set default flags in flatpak-builder which can work like below:
`flatpak-builder flags + upstream flags` + `runtime flags` + `app flags`.
The last two would be optional and probably not widely used.

src/builder-manifest.c
```diff
@@ -2834,7 +2834,16 @@ builder_manifest_finish (BuilderManifest *self,
 
       if (self->build_runtime)
         {
-          debuginfo_dir = g_file_resolve_relative_path (app_dir, "usr/lib/debug");
+          const char * libdir = builder_options_get_libdir (self->build_options, context);
```
alexlarsson: I don't think we want to modify /usr/lib/debug actually. It already contains the full path below it. For example `/usr/lib/debug/usr/lib/libglib-2.0.so.0.5200.3.debug`.

src/builder-post-process.c
```diff
@@ -377,9 +377,21 @@ builder_post_process_debuginfo (GFile          *app_dir,
...
-          real_debug_dir = g_build_filename ("/usr/lib/debug", rel_path_dir, NULL);
-          source_dir_path = g_build_filename (app_dir_path, "usr/lib/debug/source", NULL);
+          BuilderOptions * options = builder_context_get_options (context);
+          const char * libdir = builder_options_get_libdir (options, context);
+
+          if (libdir)
+            {
+              debug_dir = g_build_filename (app_dir_path, libdir, "debug", rel_path_dir, NULL);
+              real_debug_dir = g_build_filename (libdir, "debug", rel_path_dir, NULL);
+              source_dir_path = g_build_filename (app_dir_path, libdir, "debug/source", NULL);
```
alexlarsson: Same here. There only ever need to be one prefix for "where the debuginfo is stored", which is /usr/lib/debug. And then multiarch works fine because the path is stored below that.

src/builder-sdk-config.c
```diff
@@ -0,0 +1,261 @@
...
+ */
+
+#include "builder-sdk-config.h"
+
+#include <json-glib/json-glib.h>
+
+struct BuilderSdkConfig {
+  GObject       parent;
+
+  char         *prefix;
```
alexlarsson: Is there ever a need to tweak the global prefix? I'm pretty sure that won't even work due to various things being hardcoded. Can we drop this until its neeeded+tested.

alexlarsson:
Can you also add a sdk-default-override to BuilderOptions, and look up (for each key) in that (per-module and then globally) before falling back to the value from the sdk json one.

src/builder-sdk-config.c
valentindavid: It felt natural to add it when writing it. But maybe it would not work and I do not mind if I have to remove it.

src/builder-context.c
```diff
@@ -192,6 +196,7 @@ builder_context_constructed (GObject *object)
   self->cache_dir = g_file_get_child (self->state_dir, "cache");
   self->checksums_dir = g_file_get_child (self->state_dir, "checksums");
   self->ccache_dir = g_file_get_child (self->state_dir, "ccache");
+  self->sdk_defaults = g_file_get_child (self->state_dir, "sdk-defaults.json");
```
alexlarsson: Hmm, this is not right. The state dir can be shared by all kinds of builds.

valentindavid: The problem is that a build might overwrite this file. So we do not want to take it from the build directory. Do you have a better way to read a file from the SDK?

valentindavid:
> Can you also add a sdk-default-override to BuilderOptions, and look up (for each key) in that (per-module and then globally) before falling back to the value from the sdk json one.

This one is done.

src/builder-context.c
valentindavid: Would making it a directory and store them by project name work?

alexlarsson:
I modified you code a bit, new version here: https://github.com/alexlarsson/flatpak-builder/tree/sdk-defaults

valentindavid:
There is a NULL dereference. Sorry about that. Here is a fix:

https://github.com/valentindavid/flatpak-builder/commit/eeef578b8fd2dcc88b79b9318c75a7334dbed23a

Otherwise I am OK with it.

PR #143, opened 2018-05-03
[RFC] Add support for FTP sources using libcurl

larchunix:
Here is a possible implementation for downloading FTP sources using `libcurl` (MIT licensed).

@alexlarsson: Could you give me some feedback ? I would like to know if you think it is upstreamable before spending more time on it.

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -9,6 +9,7 @@ dn=$(dirname $0)
 pkg_install sudo which attr fuse \
     libubsan libasan libtsan elfutils-libelf-devel libdwarf-devel \
     elfutils git gettext-devel libappstream-glib-devel bison \
+    libcurl-devel \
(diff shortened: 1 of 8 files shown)


TingPing:
Couldn't it just be a runtime dep using the `curl` tool similar to the other source types?

alexlarsson:
Looks good to me. Although i wonder if we should just drop libsoup support altogether instead? Then we get things like HTTP2 support

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
Oh, you need to add `libcurl-devel` to `ci/build.sh`.

alexlarsson:
bot, add author to whitelist

larchunix:
bot, retest this please

TingPing:
@rh-atomic-bot r+

PR #145, opened 2018-05-03
When using --download-only skip some checks

alexlarsson:
When downloading only we will not be building, so we can skip all the checks for dependencies
and for the state-dir properties.


alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot r+

PR #147, opened 2018-05-07
Remove cmake property

barthalion:
It is deprecated since 2016-11-29, it should be enough for everyone to
move to buildsystem instead.

### DIFF
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -465,10 +465,6 @@
                     <listitem><para>Don't fix up the *.py[oc] header timestamps for ostree use.</para></listitem>
                 </varlistentry>
                 <varlistentry>
-                    <term><option>cmake</option> (boolean)</term>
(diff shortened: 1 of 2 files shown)


alexlarsson:
Well, its still used in e.g. the freedesktop json manifest. I don't see the huge gain in removing it.

barthalion:
I guess it shouldn't be marked as deprecated then? I can see new users being confused by `cmake` and `buildsystem` and every now and then we need to correct someone on Flathub to use the latter. I'll submit changes to freedesktop manifest later today.

TingPing:
It could also print a warning for a few releases first?

alexlarsson:
Yeah, lets make it print warnings.

alexlarsson:
Anyone want to make a warning print patch?

alexlarsson:
closing in favour of https://github.com/flatpak/flatpak-builder/pull/192

PR #149, opened 2018-05-09
Fix extra-data size type in manifest manpage

ptomato:
At least as of 0.10.10, this needs to be a number.


TingPing:
@rh-atomic-bot r+

PR #150, opened 2018-05-14
Add xxflags-override flags instead of sdk-default-override

alexlarsson:
This is more generic as it allows you to override at any point in the hierarchy.


alexlarsson:
@rh-atomic-bot r+

PR #151, opened 2018-05-15
Use libcurl for HTTP downloads as well

larchunix:
As discussed in #143, drop libsoup support for source download and use libcurl instead so that we can benefit of libcurl features like HTTP2 support.

By the way, it brings a fancy progress bar for downloads.

### DIFF
--- a/src/builder-context.c
+++ b/src/builder-context.c
@@ -391,13 +391,11 @@ builder_context_download_uri (BuilderContext *self,
           if (builder_download_uri (mirror_uri,
                                     dest,
                                     checksums, checksums_type,
-                                    builder_context_get_soup_session (self),
                                     builder_context_get_curl_session (self),
                                     &my_error))
             return TRUE;
 
-          if (!g_error_matches (my_error, SOUP_HTTP_ERROR, SOUP_STATUS_NOT_FOUND) &&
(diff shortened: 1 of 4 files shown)


larchunix:
@rh-atomic-bot try

larchunix:
bot, test pull request

alexlarsson:
We still use soup in src/builder-source-file.c, and there are some uses in src/builder-flatpak-utils.c which i think we can just delete.

larchunix:
> We still use soup in src/builder-source-file.c, and there are some uses in src/builder-flatpak-utils.c which i think we can just delete.

I thought the same at first but after digging more into `src/builder-source-file.c`, my understanding is that `download_uri()` function which still uses libsoup will be called only for `data:` URLs (`is_inline` boolean). As libcurl does not seems support this scheme I left this part unchanged.

For `src/builder-flatpak-utils.c`, I guess it is safe to remove the function `flatpak_download_http_uri()` which appears to be unused.

@alexlarsson: do you agree with my analysis ?

alexlarsson:
True, but we should probably rename download_uri, to download_data_uri()

alexlarsson:
Also, you need to make curl non-optional in configure.

PR #157, opened 2018-05-27
 Make libcurl return error on HTTP response >= 400 

larchunix:
I will try to make another PR to print a more useful error message when I will be less busy.

Using `CURLOPT_ERRORBUFFER` we could get `The requested URL returned error: 404 Not Found` instead of `HTTP response code said error` as we have now.


alexlarsson:
@rh-atomic-bot r+

PR #162, opened 2018-06-10
Don't prune build stages from cache with --{export,finish}-only

larchunix:
A commonly used CI workflow is to chain the following steps:

    > flatpak-builder --stop-at=<module>
    > flatpak build <commands to build module from local checkout>
    > flatpak-builder --finish-only

Unfortunately, the last step always purge all the compilation cache
created by the first one. All build steps are marked unused by default
and since they are skipped due to the option --finish-only, all of them
are always pruned by the gc function.

As a result, flatpak-builder cache becomes useless, a full compilation
is always performed.

Improve this by not cleaning unused stages when no compilation is done
by flatpak-builder (i.e: when flatpak-builder is used with --export-only
or --finish-only option).

See also: https://gitlab.gnome.org/GNOME/Initiatives/wikis/DevOps-with-Flatpak

### DIFF
--- a/src/builder-cache.c
+++ b/src/builder-cache.c
@@ -1210,6 +1210,7 @@ builder_cache_disable_lookups (BuilderCache *self)
 
 gboolean
 builder_gc (BuilderCache *self,
+            gboolean      prune_unused_stages,
             GError      **error)
 {
   gint objects_total;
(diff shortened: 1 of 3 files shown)


larchunix:
CI failure seems due to a regression in flatpak between 0.11.7 and 0.11.8.

Issue #155, opened 2018-05-24
rename-desktop-file leaves launchable entry in appdata unchanged


Alexander-Wilms:
`<launchable type="desktop-id">texstudio.desktop</launchable>`
doesn't get changed to
`<launchable type="desktop-id">org.texstudio.TeXstudio.desktop</launchable>`
when having `rename-desktop-file: texstudio.desktop` in the manifest.

Cf. https://github.com/flathub/org.texstudio.TeXstudio/issues/2

Jehan:
So for the record, we had the issue at GIMP for a long time until we finally understood it (thanks to @hughsie!). We are now patching our appdata for flatpak only so the bug is worked around for now, but it would be nice to be fixed at some point.
See: https://github.com/flathub/org.gimp.GIMP/issues/8

PR #164, opened 2018-06-13
Rewrite the <launchable> tag when using rename-desktop-file

hughsie:
This copies FlatpakXml from the flatpak project.

Fixes https://github.com/flatpak/flatpak-builder/issues/155

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -2420,35 +2420,54 @@ builder_manifest_cleanup (BuilderManifest *self,
 
           if (appdata_file != NULL)
             {
-              g_autofree char *contents;
-              const char *to_replace;
-              const char *match;
-              g_autofree char *replace_src = NULL;
-              g_autofree char *replace_dst = NULL;
+              FlatpakXml *n_id;
+              FlatpakXml *n_root;
+              FlatpakXml *n_text;
+              g_autoptr(FlatpakXml) xml_root = NULL;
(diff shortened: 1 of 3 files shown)


src/builder-manifest.c
```diff
@@ -2420,35 +2420,54 @@ builder_manifest_cleanup (BuilderManifest *self,
...
-                  g_string_append (new_contents, replace_dst);
-                  to_replace = match + strlen (replace_src);
+                  g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED, "no <component> node");
+                  return FALSE;
+                }
+              n_id = flatpak_xml_find (n_root, "id", NULL);
+              if (n_id)
+                {
+                  n_text = n_id->first_child;
+                  if (n_text && g_strcmp0 (n_text->text, self->id) != 0)
```
alexlarsson: I don't think this is quite right. There can be multiple components. This happens for instance in libreoffice which has components for each sub-app (draw, calc, writer, etc). I think we should only rewrite the ones that match the original desktop file, no?

hughsie: so more `if (n_text && g_strcmp0 (n_text->text, self->rename_desktop_file) == 0)`?

hughsie: Fixed, and commit message improved to reflect reality.

alexlarsson:
@rh-atomic-bot r+

PR #165, opened 2018-06-15
Use GInputStream for checksum computation of source files & archives

### LINKED ISSUE #141: "Failed to download sources" for files larger than ≈4 GiB
It's not possible to "download" sources larger than, approximately (few KiB less than), 4 GiB.
flatpak-builder will fail on 'Downloading sources' (though it's actually a local file).

> Downloading sources
> Failed to download sources: module BROKEN: Error reading from file: Bad address

With "--disable-download" building works.

Fedora 27, x86_64, 16 GiB RAM
flatpak-builder 0.10.10

cxrvh:
Using g_load_contents () for the checksum computation uses a lot of memory
(for large files) and actually fails for files larger than 4 GiB. Instead,
this implementation uses GInputStream to read the file in small buffers
and feeds the GChecksum manually.

Fixes #141.

I'm passing the GFile pointer in builder_verify_checksums () because I haven't found out how to clone a GObject or else I'd just pass GInputStream pointer.

### DIFF
--- a/src/builder-source-archive.c
+++ b/src/builder-source-archive.c
@@ -358,21 +358,10 @@ builder_source_archive_download (BuilderSource  *source,
 
   if (g_file_query_exists (file, NULL))
     {
-      if (is_local && checksums[0]  != NULL)
-        {
-          g_autofree char *data = NULL;
-          gsize len;
-
-          if (!g_file_load_contents (file, NULL, &data, &len, NULL, error))
-            return FALSE;
-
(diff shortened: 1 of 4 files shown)


alexlarsson:
bot, add author to whitelist

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot retry

PR #167, opened 2018-06-21
install-deps: Always pass -y when installing

alexlarsson:
The prompts don't currently work when we shell out to
flatpak, so this is broken without -y.


alexlarsson:
@rh-atomic-bot r+

Issue #170, opened 2018-06-25
--install broken


eszlari:
The commit https://github.com/flatpak/flatpak-builder/commit/6d9f874c12ea3e742f15fe014138884da9130e08 is supposed to fix `--install`, but it doesn't:
```
$ flatpak-builder --version
flatpak-builder 0.99.1
$ flatpak-builder --ccache --install --force-clean .app-dir org.example.App.yaml
[...]
Is this ok [y/n]: n
```

alexlarsson:
Nah, that fix is for --install-deps-from= and we need a similiar one for --install

PR #171, opened 2018-06-25
Fix --install with latest flatpak by passing -y

alexlarsson:
These days flatpak install will always prompt for input,
so we pass -y to it because otherwise (due to non-inherited
stdin) we auto-answer no.

Fixes https://github.com/flatpak/flatpak-builder/issues/170


alexlarsson:
@rh-atomic-bot r+

PR #174, opened 2018-06-26
builder: Initialise git tree when using "use-git" for patches

hadess:
When applying patches using git, git apply will ignore non-existent
patches that apply to non-existent files. As the vanilla source we use
to apply those patches might not be a git tree (eg. a tarball), make
sure that the tree is initialised before apply those git patches.

Fixes "Skipped patch ..." messages appearing in builds.

### DIFF
--- a/src/builder-source-patch.c
+++ b/src/builder-source-patch.c
@@ -39,6 +39,7 @@ struct BuilderSourcePatch
   char         *path;
   guint         strip_components;
   gboolean      use_git;
+  gboolean      use_git_am;
   char        **options;
 };
 
@@ -55,6 +56,7 @@ enum {
   PROP_STRIP_COMPONENTS,
   PROP_USE_GIT,
   PROP_OPTIONS,
+  PROP_USE_GIT_AM,
   LAST_PROP
 };
 
@@ -95,6 +97,10 @@ builder_source_patch_get_property (GObject    *object,
       g_value_set_boxed (value, self->options);
       break;
 
+    case PROP_USE_GIT_AM:
+      g_value_set_boolean (value, self->use_git_am);
+      break;
+
     default:
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
     }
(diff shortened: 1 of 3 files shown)


src/builder-source-patch.c
```diff
@@ -186,6 +186,46 @@ builder_source_patch_download (BuilderSource  *source,
...
+static gboolean
+ensure_git (GFile *dir)
+{
+  GFile *git_dir;
+  gboolean ret;
+  char *basename;
+
+  git_dir = g_file_get_child (dir, ".git");
+  ret = g_file_query_exists (git_dir, NULL);
+  g_object_unref (git_dir);
```
alexlarsson: I think this should try "git status" instead, because it might be run in a subdirectory of the directory with a .git in it.

hadess: Sorry, don't understand. The patch would be applied to a sub-dir that's a git repo, not at the top-level?

hadess:
I've added another patch to use git-am to apply patches, because git-apply completely mucks file renames. I'm not certain whether this would warrant another option, or will break application of some patches. Do we have a test suite for flatpak-builder?

src/builder-source-patch.c
alexlarsson: Say the toplevel is a git checkout and you want to apply a patch to a subdirectory only (dest=subdir in the source), then it will look for subdir/.git which doesn't exist, rather than the toplevel .git which does, and then wrongly do git init.

alexlarsson:
use-git was originally added (on your request i believe) to handle patches of binaries. I'm not sure switching it to git am is a good idea, because i belive that requires the patches to be in git format-patch format. So, we should probably add a use-git-am option for this.

hadess:
I added 2 new options, one to initialise a decompressed archive with git, and another one to apply patches with git-am. Works in my local testing with https://github.com/flathub/flathub/pull/443

alexlarsson:
@rh-atomic-bot r+

hadess:
The `git-init` patch isn't quite good enough:
https://flathub.org/builds/#/builders/41/builds/231

We also need to pass some data for email and name.

PR #178, opened 2018-07-02
Support alternative builds

alexlarsson:
This allows you to specify alternative builds in one manifest file.
You specify the alternative on the commandline like:
  flatpak-builder --alt=NAME ...

And this then builds the manifest like normally, but with a different
cache base-name (so that both alternatives can be cached at the
same time). However, using the new "alt" option in BuildOption
and [only|skip]-alts options in modules and sources you can
affect what gets built and how. This is very similar to how
things can already be per-arch optional.

### DIFF
--- a/doc/flatpak-builder.xml
+++ b/doc/flatpak-builder.xml
@@ -349,6 +349,16 @@
             </varlistentry>
 
             <varlistentry>
+                <term><option>--alt=ALTNAME</option></term>
+
+                <listitem><para>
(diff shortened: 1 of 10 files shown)


alexlarsson:
Hmm, maybe we should make this a multi-argument thing instead. Like --enable=feature1 --enable=feature2, and then enable-if: ["feature1"].

matthiasclasen:
--enable/disable feature sounds more future-proof to me

Issue #172, opened 2018-06-25
Provide FLATPAK_BUILDER_MOD_SRC


eszlari:
It would be nice to have a variable to refer to the location of the source tree of the module that is built, instead of having to use `/run/build/MODULE_NAME`.

alexlarsson:
Yeah, that seems useful. Anyone want to make a patch?

PR #180, opened 2018-07-05
Set FLATPAK_BUILDER_BUILDDIR in env when building module

alexlarsson:
This is set to the /run/build/$module directory name so that you
can easily refer to it in the build.

Fixes https://github.com/flatpak/flatpak-builder/issues/172


alexlarsson:
@rh-atomic-bot r+

PR #181, opened 2018-07-10
rename-desktop-file: Add X-Flatpak-RenamedFrom key

alexlarsson:
This way desktop environment can handle the case
where a desktop file was renamed and avoid showing
two icons.


alexlarsson:
@rh-atomic-bot r+

PR #183, opened 2018-07-13
builder: Ensure committer is set when running git-commit

hadess:
Otherwise buildbots, and other non-interactive/clean-room builds might
not have git committer information set, and make it impossible to
successfully commit, and git erroring out with:
*** Please tell me who you are.

This should fix https://github.com/flathub/flathub/pull/443#issuecomment-402062013

### DIFF
--- a/src/builder-main.c
+++ b/src/builder-main.c
@@ -299,6 +299,70 @@ do_install (BuilderContext *build_context,
   return TRUE;
 }
 
+static gboolean
+git (char   **output,
+     GError **error,
+     ...)
+{
+  gboolean res;
+  va_list ap;
+
+  if (output != NULL)
+    *output = NULL;
+
+  va_start (ap, error);
+  res = flatpak_spawn (NULL, output, 0, error, "git", ap);
+  va_end (ap);
+
+  if (output != NULL &&
+      (*output != NULL && *output[0] == '\0'))
+    {
+      g_free (*output);
+      *output = NULL;
+    }
+
+  return res;
+}
+
+static char *
+trim_linefeed (char *str)
+{
+  guint len;
+
+  g_return_val_if_fail (str != NULL, NULL);
+
+  len = strlen (str);
+  str[len] = '\0';
+
+  return str;
+}
+
+static void
+git_init_email (void)
+{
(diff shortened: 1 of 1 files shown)


hadess:
Fully tested, but uglier than expected.

src/builder-source-archive.c
```diff
@@ -638,18 +638,42 @@ static gboolean
...
-  return TRUE;
+  /* Have an email for author and committer */
+  if (!git (dir, &email, NULL, "config", "--get", "user.email", NULL) ||
+      email == NULL)
+    email = g_strdup ("flatpak-builder-commit@flatpak.org");
+  else
+    email = trim_linefeed (email);
+  g_setenv ("GIT_AUTHOR_EMAIL", email, FALSE);
+  g_setenv ("GIT_COMMITTER_EMAIL", email, FALSE);
+  g_free (email);
```
alexlarsson: Overriding the environment globally like this feels a bit weird, as it will affect all processes that f-b later spawns.
Either we should pass the env vars specifically only to the child, or we should do this entire dance up-front in main() so that it affect all git calls.

alexlarsson:
I think just forcing GIT_AUTHOR_EMAIL and GIT_COMMITTER_EMAIL to be set in main() is the right approach. Then we're guaranteeing this to work for all git invocations, and we make the result more reproducible than sometimes using the real authors name. No created commits from flatpak-builder are persistent, except possibly flatpak-builder --run, so we should not set it there.

hadess:
> I think just forcing GIT_AUTHOR_EMAIL and GIT_COMMITTER_EMAIL to be set in main() is the right approach. Then we're guaranteeing this to work for all git invocations, and we make the result more reproducible than sometimes using the real authors name. No created commits from flatpak-builder are persistent, except possibly flatpak-builder --run, so we should not set it there.

If I do that, we don't have a `GIT_AUTHOR_NAME` that corresponds to the package name, but that's not really a big deal.

alexlarsson:
Yeah, that doesn't seem as important.

alexlarsson:
@rh-atomic-bot r+

PR #187, opened 2018-07-26
manpage: More info about extension metadata.

ptomato:
This refers to the flatpak-metadata manpage for more info on the extension
metadata properties, as well as clarifying a key piece of information about
the directory property.

See flatpak#1341.


TingPing:
@rh-atomic-bot r+

PR #190, opened 2018-08-01
module: Respect "no-parallel-make" with ninja

hadess:
ninja defaults to running in a highly parallel mode, but some bugs in
meson, such as:
https://github.com/mesonbuild/meson/issues/1994
require building serially.

This change makes the cmake-ninja and meson buildsystems respect
"no-parallel-make" during building.


TingPing:
@rh-atomic-bot r+

PR #191, opened 2018-08-03
Fix inline "data:" URLs

rshuler:
Once we detect that a URL is inline text, we must not try to
actually download it.


mwleeds:
@rh-atomic-bot r+ f6faafe

PR #192, opened 2018-08-07
Warn about using deprecated cmake property.

alexlarsson:


src/builder-module.c
```diff
@@ -341,6 +341,8 @@ builder_module_set_property (GObject      *object,
 
     case PROP_CMAKE:
       self->cmake = g_value_get_boolean (value);
+      if (self->cmake)
+        g_printerr ("The cmake module property is deprecated, use buildsystem cmake instead.\n");
```
TingPing: Maybe mention `cmake-ninja` also?

alexlarsson: I dunno how important that is for the warning.

TingPing: Well it just avoids wasting everybodies time IMO:

- User sees new warning, follows warning and uses `buildsystem: cmake`
- Somebody else points out `cmake-ninja` exists now and is just better
- User updates it a second time

alexlarsson: Changing to cmake-ninja is not just a deprecation avoidance though, but could cause build issues. I guess its worth mentioning though.

alexlarsson:
@rh-atomic-bot r+

Issue #184, opened 2018-07-19
X-Flatpak-RenamedFrom needs to be a list


ramcq:
#181 is great and all... but... is there some way we can format / spec this as a ;-separated list and allow multiple old names to be provided? We have some cases in Flathub where we need to rename a desktop file a 2nd time: LibreOffice's .desktop files are currently shipping with names that aren't real sub-IDs of org.libreoffice.LibreOffice so we need to actually rename them again! And we have app IDs which want/need to change. Endless has an X-Endless-Alias key that does the same as this... so it's cool we have an upstream version, but it also has the same bug of assuming renaming is only once, so I need to add some lame X-Endless-Aliases key to handle the twice-renamed things, and support both for back compatibility. Can we fix it here from the outset in the upstream version...? :)

See https://github.com/flathub/org.libreoffice.LibreOffice/issues/36#issuecomment-389141781

matthiasclasen:
On Thu, Jul 19, 2018 at 7:02 PM, Robert McQueen <notifications@github.com>
wrote:

> #181 <https://github.com/flatpak/flatpak-builder/pull/181> is great and
> all... but... is there some way we can format / spec this as a ;-separated
> list and allow multiple old names to be provided? We have some cases in
> Flathub where we need to rename a desktop file a 2nd time: LibreOffice's
> .desktop files are currently shipping with names that aren't real sub-IDs
> of org.libreoffice.LibreOffice so we need to actually rename them again!
> And we have app IDs which want/need to change. Endless has an
> X-Endless-Alias key that does the same as this... so it's cool we have an
> upstream version, but it also has the same bug of assuming renaming is only
> once, so I need to add some lame X-Endless-Aliases key to handle the
> twice-renamed things, and support both for back compatibility. Can we fix
> it here from the outset in the upstream version...? :)
>
> See flathub/org.libreoffice.LibreOffice#36 (comment)
> <https://github.com/flathub/org.libreoffice.LibreOffice/issues/36#issuecomment-389141781>
>
Turning it into a list is a little annoying, since lists in desktop files
are ;-terminated, not ;-separated. So this will change the syntax even for
single renames to require a ; at the end.

ramcq:
That's what I was thinking, this is an unstable version still, so best to just change it ASAP, otherwise later we'd have to add a list version anyway.

PR #193, opened 2018-08-07
Make X-Flatpak-RenamedFrom a list so that we can handle multiple renames

alexlarsson:
Fixes https://github.com/flatpak/flatpak-builder/issues/184


alexlarsson:
@rh-atomic-bot r+

Issue #130, opened 2018-04-21
--skip-if-unchanged should be an arch specific check.


Jehan:
I realize that when I run flatpak-builder with --skip-if-unchanged, for 2 archs consecutively, the second build is not done, returning with:

> No changes to manifest, skipping

Steps:

- Change your manifest
- Run `flatpak-builder --skip-if-unchanged --arch=x86_64 mymanifest.json`. It works fine.
- Run `flatpak-builder --skip-if-unchanged --arch=i386 mymanifest.json`

The second command returns immediately telling us the manifest has no changes. I assume that is because it has just run the same manifest, but that was for another architecture! Therefore it should actually build the flatpak.
Of course this is no problem when your build system is on several servers (one per arch for instance), but that is a problem with a single build server and when you want to optimize a bit your build time by not rebuilding unchanged manifests.

PR #197, opened 2018-08-08
Take arch into account for --skip-if-unchanged

alexlarsson:
This fixes https://github.com/flatpak/flatpak-builder/issues/130


alexlarsson:
@rh-atomic-bot r+

PR #198, opened 2018-08-08
main: ensure mirror URLs are /-terminated

ramcq:
The mirror URL is composed with the file we want using soup_uri_new_with_base,
which means a mirror path that is not /-terminated will have it's last element
stripped off. When parsing the list of sources URI arguments, add a / if needed
before the Soup base URI object is created.


alexlarsson:
@rh-atomic-bot r+

PR #203, opened 2018-08-15
Keep x-* properties in manifest.json

alexlarsson:
This add custom (de)serialization code for the source, module and manifest
objects so that properties starting with "x-" are kep and then put back
in the manifest.

We also add a checksum of the manifest to the "finish" phase so that
if you change them the manifest is re-generated.


TingPing:
Looks good and works for my usage.

@rh-atomic-bot r+

Issue #214, opened 2018-09-19
Unnecessary rebuilds for modules containing a + in their name


kirbyfan64:
Example YAML file:

```yaml
app-id: com.test.Test
runtime: org.freedesktop.Platform
runtime-version: 18.08
sdk: org.freedesktop.Sdk
modules:
  - name: libxpp
    buildsystem: simple
    build-commands:
      - 'echo "Build libxpp"'
  - name: libx++
    buildsystem: simple
    build-commands:
      - 'echo "Build libx++"'
  - name: test
    buildsystem: simple
    build-commands:
      - 'echo "Build test"'
```

- When you run `flatpak-builder` the first time, it (of course) runs a full build.
- Run it again, and it reuses the cache.
- Run it again, and it rebuilds `libx++`.
- Run it again, and it reuses the cache.
- Run it again, and it rebuilds `libx++`
- Run it again, and it reuses the cache.
- ...

Basically, every other build results in rebuilding libx++.

alexlarsson:
Hmm, can you try quoting the name?

kirbyfan64:
I'll try, but FWIW I've also gotten this to occur with JSON, where the names have to be quoted.

alexlarsson:
We use the module name as a ostree ref in the cache. I wonder if + is handled specially by ostree when part of a ref name.

alexlarsson:
Hmm, here is the cache:
```
$ flatpak-builder --force-clean app cache-plus.yaml 
first build...
$ ls .flatpak-builder/cache/refs/heads/x86_64-cache-plus.yaml/
build-libx2b2b  build-libxpp  build-test  cleanup  finish  init
$ flatpak-builder --force-clean app cache-plus.yaml 
used cache...
$ ls .flatpak-builder/cache/refs/heads/x86_64-cache-plus.yaml/
build-libxpp  build-test  cleanup  finish  init
$ flatpak-builder --force-clean app cache-plus.yaml 
$ ls .flatpak-builder/cache/refs/heads/x86_64-cache-plus.yaml/
build-libx2b2b  build-libxpp  build-test  cleanup  finish  init
```
So, when building things we create the build-libx2b2b ref, but when reusing it we delete it. I think that is because ostree automatically escapes the + in the ref, but the "unused refs GC" we do after a build did not.

PR #215, opened 2018-09-21
cache: Make sure we escape stage names when we mark them as done

alexlarsson:
We need to to this, because they are also escaped as refs in the cache.

Fixes https://github.com/flatpak/flatpak-builder/issues/214


alexlarsson:
@rh-atomic-bot r+

PR #217, opened 2018-09-26
Run screenshot mirroring in sandbox

alexlarsson:
The old version of appstream-glib on the flathub builders are causing problems, and we already rely on appstream-glib in the runtime for appstream-compose, so we might as well run the screenshot mirroring in the sandbox too.

Also, this fixes an issue where the rewrite of the xml file causes issues with the f-b cache repo.


alexlarsson:
@rh-atomic-bot r+

PR #221, opened 2018-10-15
Print a warning when a YAML document has a float-like string value

kirbyfan64:
This has personally happened to me a few different times, causing quite a bit of confusion!


alexlarsson:
@rh-atomic-bot r+

Issue #247, opened 2018-12-27
Questions re .pyc cleanup step


lazka:
## Linux distribution and version

Debian sid

## Flatpak-builder version

flatpak-builder 1.0.1

## Description of the problem

As things are now flatpak-builder removes all .pyc files that get created during the build with distutils/setuptools. Example: https://flathub.org/builds/#/builders/16/builds/721/steps/5/logs/stdio
Looking at the code it looks like it shouldn't, but it also hardcodes the mtime to bytes 4-8 in fixup_python_time_stamp() which is no longer the case with Python 3.7, so not sure.

A related issue for the fdo-sdk is discussed here: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/issues/554 where the stdlib .pyc files have a wrong mtime. The current consensus there is to move Python to hash based verification (instead of mtime) by default which is supported since Python 3.7.

So, assuming this gets solved in the fdo-sdk, and the current broken state of the pyc cleanup code I'm wondering if it shouldn't be removed.

In case someone wonders why having this working would be useful: For Quod Libet having .pyc for its own source files (excluding stdlib) saves 500ms on start.

nanonyme:
I don't think it in general can be removed but it would be nice if we could gradually deprecate and kill the functionality through opt-out. FD.o 1.6 has Python 2 in Platform, FD.o 18.08 and derivatives has Python 2 only in SDK and FD.o 19.08 will not have Python 2 at all.

nanonyme:
Note that flatpak-builder should most likely set SOURCE_DATE_EPOCH to something reasonable so our Python 3.7 uses CHECKED_HASH bytecode by default.

nanonyme:
In runtime we're hardcoding it into a constant like this https://buildstream.gitlab.io/buildstream/format_project.html#builtin-defaults

nanonyme:
Magic byte contains Python version. It might be best if this entire logic was disabled on 3.7 and newer. Ping @alexlarsson

alexlarsson:
The mtime munging is definately needed on older versions, because ostree itself will modify the mtime of things, so the embedded mtimes will be wrong. However, if python changed we need to update this.

nanonyme:
@alexlarsson the core thing here is format changed and bytecode may or may not have timestamps at all so you can't simply write without parsing the header

alexlarsson:
Yeah, the code needs updating. But we can't just disable it.

nanonyme:
[...]

PR #255, opened 2019-01-11
Set SOURCE_DATE_EPOCH to the mtime of the manifest

alexlarsson:
This will help make reproducible builds.

As mentioned in https://github.com/flatpak/flatpak-builder/issues/247#issuecomment-450585337


alexlarsson:
@rh-atomic-bot r+

PR #256, opened 2019-01-14
post-process: Handle python 3.7 header formats

### LINKED ISSUE #247: Questions re .pyc cleanup step

alexlarsson:
Python 3.7 adds a flags field which modified the offset of the mtime,
as well as adds a new non-mtime based check mode.

Fixes #247

### DIFF
--- a/src/builder-post-process.c
+++ b/src/builder-post-process.c
@@ -112,6 +112,9 @@ invalidate_old_python_compiled (const char *path,
   return TRUE;
 }
 
+/* We need to read at least 12 bytes to get both magic + optional flags + mtime */
+#define PYTHON_HEADER_SIZE 12
+
 static gboolean
 fixup_python_time_stamp (const char *path,
                          const char *rel_path,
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

PR #257, opened 2019-01-14
Allow unsetting SOURCE_DATE_EPOCH and other env vars

alexlarsson:
As per https://github.com/flatpak/flatpak-builder/issues/247#issuecomment-453937152, the new SOURCE_DATE_EPOCH support can cause problems for some apps, and we need to be able to disable it.

With this change, QuodLibet build with this change:

```
diff --git a/io.github.quodlibet.QuodLibet.yaml b/io.github.quodlibet.QuodLibet.yaml
index 3252741..147fef6 100644
--- a/io.github.quodlibet.QuodLibet.yaml
+++ b/io.github.quodlibet.QuodLibet.yaml
@@ -118,6 +118,9 @@ modules:
         url: https://gstreamer.freedesktop.org/src/gst-plugins-bad/gst-plugins-bad-1.14.4.tar.xz
         sha256: 910b4e0e2e897e8b6d06767af1779d70057c309f67292f485ff988d087aa0de5
   - name: quodlibet
+    build-options:
+      env:
+        SOURCE_DATE_EPOCH: null
     buildsystem: simple
     subdir: quodlibet
     build-commands:
```

### DIFF
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -340,7 +340,7 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>env</option> (object)</term>
-                    <listitem><para>This is a dictionary defining environment variables to be set during the build. Elements in this override the properties that set the environment, like cflags and ldflags.</para></listitem>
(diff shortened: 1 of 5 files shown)


alexlarsson:
@rh-atomic-bot r+

PR #258, opened 2019-01-15
docs: Clarify that command: is only for flatpak run

matthiasclasen:
This may not otherwise be obvious.


alexlarsson:
@rh-atomic-bot r+

PR #264, opened 2019-01-25
manifest: Add default-branch key

alexlarsson:
This is similar to branch and used if branch is not set. However, this
key (as opposed to branch) is overridden by the --default-branch option.

The idea is that most apps in flathub would use default-branch=stable, so
that a standart flatpak-builder command will build a "stable" release.
However, when building a test build we can use --default-branch=test to
override that. Then special cases like theme extensions and other things
that require a specific branch value to work at all can use branch="1.0".

### DIFF
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -58,7 +58,24 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>branch</option> (string)</term>
-                    <listitem><para>The branch of the application, defaults to master.</para></listitem>
+                    <listitem><para>The branch to use when exporting
+                    the application. If this is unset the defaults
+                    come from the default-branch option.</para>
+                    <para>This key overrides both the default-branch
+                    key, and the --default-branch commandline
+                    option. Unless you need a very specific branchname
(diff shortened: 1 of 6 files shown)


alexlarsson:
@rh-atomic-bot r+

matthiasclasen:
Why do we need both branch and default-branch ? Does that buy us anything over having 
branch overridable with `--branch` ? Is the point that branch is _not_ overridable ?

alexlarsson:
The goal is this:
 * If you check out the main branch of a flathub git repo and run flatpak-builder app.json (with no special options) it will build a "stable" branch
 * Builds in normal flathub will also build a "stable" branch
 * Test builds in flathub will build a "test" branch
 * Non-applications, like icon themes or similar things have very specific needs for the branch name. (Icon themes must have branch 1.0, gtk theme 3.24, etc.)

The last goal means we can't always pass --branch stable that overrides the branch from the manifest, therefor flathub passed --default-branch=stable. This works fine only if you don't specify *any* branch in the repo, but in that case a local build with no --default-branch argument will build a "master" branch, which is not great.

PR #266, opened 2019-01-30
manifest: use the sdk branch from the sdk ref specified by the user

MatMaul:
Allow install_extension_deps to take the sdk branch as param as a side effect.

This is to accommodate this kind of app extension building:
https://github.com/MatMaul/flathub/blob/org.videolan.VLC.Plugin.bluray/org.videolan.VLC.Plugin.bluray.json

Needs https://github.com/flatpak/flatpak/pull/2664.

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -3660,6 +3660,7 @@ static gboolean
 builder_manifest_install_extension_deps (BuilderManifest *self,
                                          BuilderContext  *context,
                                          const char *runtime,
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

MatMaul:
Great thanks.
It would be nice if you can cut a release of flatpak-builder when you release flatpak 1.2.1, I need all of that to move on with the VLC Blu-ray plugin.

PR #269, opened 2019-02-06
Fixup error messages

alexlarsson:
We were printing argv[3] which is wrong (and typically null).
Instead set the error message where it happens and we know
what we were trying to do.


alexlarsson:
@rh-atomic-bot r+

PR #270, opened 2019-02-08
Fix handling of sdk default compiler options

alexlarsson:
It turns out that newer flatpaks changed the output of flatpak info
and we relied on parsing that due to the lack of --show-location option
to flatpak info. However, that has been added since flatpak 0.11.8, so
just use that instead of parsing it.


alexlarsson:
@rh-atomic-bot r+

PR #277, opened 2019-04-04
Update platform creation to ensure good mtimes and better cacheing

alexlarsson:
This splits the platform creation into 3 parts:
 * base - create the initial directory based on the parent platform
 * prepare - run prepare commands and apply all changes
 * cleanup - apply cleanups and cleanup commands

This has cacheing advantages in that prepare_commands and cleanup changes
only cause the minimal amount of rebuilds.

Additionally, it ensures that the mtimes are zeroed out (from the
previous checkout) both when the prepare and cleanup commands are run.
This is actually important, since these often generate caches (for
example fontconfig ones) which rely on zeroed mtimes so they match
what will be deployed.

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -1869,20 +1869,15 @@ builder_manifest_checksum_for_bundle_sources (BuilderManifest *self,
 }
 
 static void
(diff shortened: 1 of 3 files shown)


alexlarsson:
@rh-atomic-bot r+

PR #281, opened 2019-04-19
utils: force exit host commands when flatpak-builder exits

chergert:
If we are spawning applications on the host using the Development service,
then we want those commands to exit when the flatpak-builder process
exits, as can happen from Ctrl^C or kill().

### DIFF
--- a/src/builder-flatpak-utils.h
+++ b/src/builder-flatpak-utils.h
@@ -33,6 +33,7 @@
 
 typedef enum {
   FLATPAK_HOST_COMMAND_FLAGS_CLEAR_ENV = 1 << 0,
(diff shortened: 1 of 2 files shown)


alexlarsson:
Hmm, this is a flatpak >= 1.2 flag, and previous versions will fail with G_DBUS_ERROR_INVALID_ARGS if we set it, so we should detect that and re-try.

alexlarsson:
Its a bit lame that we added that flag but did not bump the portal version nr...

chergert:
Updated to check for invalid-args only once and retry without `WATCH_BUS`.

alexlarsson:
@rh-atomic-bot r+

alexlarsson:
@rh-atomic-bot retry

PR #284, opened 2019-05-12
Support multiple instances of install-deps-from

### LINKED ISSUE #278: Support multiple sources for installing dependencies
Its useful to install dependencies from multiple location, for example the main sdk from one place, but an sdk extension from a different one. So, we should support multiple instances of --install-deps-from=

nanonyme:
Fixes #278

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -3834,6 +3834,7 @@ builder_manifest_install_dep (BuilderManifest *self,
 {
   g_autofree char *ref = NULL;
   g_autofree char *commit = NULL;
+  g_autoptr(GError) first_error = NULL;
 
   if (version == NULL)
     version = builder_manifest_get_runtime_version (self);
@@ -3869,17 +3870,22 @@ builder_manifest_install_dep (BuilderManifest *self,
 	  if (builder_manifest_install_single_dep (ref, remote, opt_user, opt_installation,
                                                    &current_error))
 	  {
-	    current_error = g_steal_pointer(error);
 	    return TRUE;
 	  }
-	  else if (*error == NULL)
-	    {
-	      *error = g_steal_pointer(&current_error);
-	    }
-	  if ((*error)->domain != G_SPAWN_EXIT_ERROR)
-	    return FALSE;
+	  else {
+	    gboolean fatal_error = current_error->domain != G_SPAWN_EXIT_ERROR;
+	    if (first_error == NULL)
+	      {
+	        first_error = g_steal_pointer(&current_error);
+	      }
+	    if (fatal_error)
+	      {
+		break;
+	      }
+	  }
 	}
     }
+  *error = g_steal_pointer(&first_error);
   return FALSE;
 }
 
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -3754,44 +3754,53 @@ builder_manifest_show_deps (BuilderManifest *self,
 }
 
 static gboolean
-builder_manifest_install_dep (BuilderManifest *self,
-                              BuilderContext  *context,
-                              const char *remote,
-                              gboolean opt_user,
-                              const char *opt_installation,
-                              const char *runtime,
-                              const char *version,
-                              gboolean opt_yes,
-                              GError         **error)
+builder_manifest_install_single_dep (const char *ref,
+				     const char *remote,
+				     gboolean opt_user,
+				     const char *opt_installation,
+				     GError **error)
 {
-  g_autofree char *ref = NULL;
-  g_autofree char *commit = NULL;
   g_autoptr(GPtrArray) args = NULL;
-  gboolean installed;
-
-  if (version == NULL)
-    version = builder_manifest_get_runtime_version (self);
-
-  ref = flatpak_build_untyped_ref (runtime,
-                                   version,
-                                   builder_context_get_arch (context));
-
-  commit = flatpak_info (opt_user, opt_installation, "--show-commit", ref, NULL);
-  installed = (commit != NULL);
-
   args = g_ptr_array_new_with_free_func (g_free);
   g_ptr_array_add (args, g_strdup ("flatpak"));
   add_installation_args (args, opt_user, opt_installation);
-  if (installed)
+
+  g_ptr_array_add (args, g_strdup ("install"));
+  g_ptr_array_add (args, g_strdup (remote));
+  
+  g_ptr_array_add (args, g_strdup ("-y"));
+  if (flatpak_version_check (1, 2, 0))
+    g_ptr_array_add (args, g_strdup ("--noninteractive"));
+
+  g_ptr_array_add (args, g_strdup (ref));
+  g_ptr_array_add (args, NULL);
+
+  if (!builder_maybe_host_spawnv (NULL, NULL, 0, error, (const char * const *)args->pdata))
     {
-      g_ptr_array_add (args, g_strdup ("update"));
-      g_ptr_array_add (args, g_strdup ("--subpath="));
+      g_autofree char *commandline = flatpak_quote_argv ((const char **)args->pdata);
+      g_prefix_error (error, "running `%s`: ", commandline);
+      return FALSE;
     }
   else
     {
-      g_ptr_array_add (args, g_strdup ("install"));
-      g_ptr_array_add (args, g_strdup (remote));
+      return TRUE;
     }
+}
+
+
+static gboolean
+builder_manifest_update_single_dep (const char *ref,
+				     gboolean opt_user,
+				     const char *opt_installation,
+				     GError **error)
+{
+  g_autoptr(GPtrArray) args = NULL;
+  args = g_ptr_array_new_with_free_func (g_free);
+  g_ptr_array_add (args, g_strdup ("flatpak"));
+  add_installation_args (args, opt_user, opt_installation);
+
+  g_ptr_array_add (args, g_strdup ("update"));
+  g_ptr_array_add (args, g_strdup ("--subpath="));
 
   g_ptr_array_add (args, g_strdup ("-y"));
   if (flatpak_version_check (1, 2, 0))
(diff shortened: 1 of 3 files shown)


nanonyme:
Tested locally through

flatpak-builder --user --install-deps-from fdo-sdk-test-repo --install-deps-from=flathub --repo=steam --force-clean build-dir com.valvesoftware.Steam.yml

src/builder-manifest.c
```diff
@@ -3756,58 +3756,65 @@ builder_manifest_show_deps (BuilderManifest *self,
...
-
-  if (!builder_maybe_host_spawnv (NULL, NULL, 0, error, (const char * const *)args->pdata))
-    {
-      g_autofree char *commandline = flatpak_quote_argv ((const char **)args->pdata);
-      g_prefix_error (error, "running `%s`: ", commandline);
-      return FALSE;
-    }
+{
+  const char *remote = *remotes;
+  while (remote != NULL) {
```
alexlarsson: Please use a for loop here, and brackets on the next line

src/builder-manifest.c
```diff
@@ -3756,58 +3756,65 @@ builder_manifest_show_deps (BuilderManifest *self,
...
+    if (flatpak_version_check (1, 2, 0))
+      g_ptr_array_add (args, g_strdup ("--noninteractive"));
+
+    g_ptr_array_add (args, g_strdup (ref));
+    g_ptr_array_add (args, NULL);
+
+    if (!builder_maybe_host_spawnv (NULL, NULL, 0, error, (const char * const *)args->pdata))
+      {
+	g_autofree char *commandline = flatpak_quote_argv ((const char **)args->pdata);
+	g_prefix_error (error, "running `%s`: ", commandline);
```
alexlarsson: We need to handle errors and looping better. For instance, on an update we should never try again with another remote, because we didn't even use the remote argument. 

Then, if we get a error other than the flatpak command exited with exit code != 0 then we should not loop (i.e. if we have some internal error kind of thing). 

But, if we get an exit code != 0 on an install, we have to assume that it printed the error already. And, if we have more that one remote such errors are going to be expected which is not obvious to the user. So, in the multi-remote case we should print something like "Trying install from remote %s" or something, and then the error should only be set at the last remote failure. (The above is actually problematic in that it sets error multiple times.)

nanonyme: I don't get the difference between install and update. Maybe I missed some flow, it looked initially like the check in main prevented entering these install functions if remote is not given. Will review the code again.

Yeah, I guess logging needs improving.

alexlarsson: The CLI syntax for install is "$remote $id", whereas for update it is just "$id" and always pulls from previous origin.

nanonyme: I'm still confused. This is flatpak-builder, I can't find any explicit update command.

nanonyme: Ah. So we call flatpak to install. Fair enough.

nanonyme: I'll try to refactor these flows to separate internal functions , it should help help.

nanonyme: Should error only be set once? You can't know beforehand which remote is supposed to have the ref.

nanonyme: @alexlarsson maybe we should be more generic with the error anyway? Looked like process command line is already outputted on debug level.

nanonyme: Output is now clearer. I'm still trying to research how the error handling should work.

nanonyme: Eg the problem is not just that error was set by us repeatedly. Spawning the child process also sets the error so currently the code still bugs if there's 2+ first remotes that don't have the ref.

src/builder-manifest.c
```diff
@@ -3754,41 +3754,25 @@ builder_manifest_show_deps (BuilderManifest *self,
...
-  if (installed)
+  if (remote == NULL)
     {
+      g_print("Updating %s\n", ref);
       g_ptr_array_add (args, g_strdup ("update"));
       g_ptr_array_add (args, g_strdup ("--subpath="));
     }
   else
     {
+      g_print("Trying to install %s from %s\n", ref, remote);
```
alexlarsson: I think we should move this print to builder_manifest_install_dep() and skip the "trying" prefix if there is only one remote.
So, it'll be like for the standard case the output would look something like:
```
Installing $ref
... output, including an error
There was an error installing the dependency $dep
exit(1)
```

But for the multi-remote case:
```
Trying to installing $ref from $remote1
... output, including an error
Trying to installing $ref from $remote2
... output, including an error
There was an error installing the dependency $dep
exit(1)
```

src/builder-manifest.c
```diff
@@ -3754,41 +3754,25 @@ builder_manifest_show_deps (BuilderManifest *self,
...
-                                   builder_context_get_arch (context));
-
-  commit = flatpak_info (opt_user, opt_installation, "--show-commit", ref, NULL);
-  installed = (commit != NULL);
-
   args = g_ptr_array_new_with_free_func (g_free);
   g_ptr_array_add (args, g_strdup ("flatpak"));
   add_installation_args (args, opt_user, opt_installation);
-  if (installed)
+  if (remote == NULL)
```
alexlarsson: This function is really small now, and it would probably be clearer if it was split into two, one called update_single_dep(), not taking a remote arg, plus this which would always take a remote arg.

Having remote == NULL mean update is kind of unexpected.

src/builder-manifest.c
```diff
@@ -3802,12 +3786,57 @@ builder_manifest_install_dep (BuilderManifest *self,
 
   if (!builder_maybe_host_spawnv (NULL, NULL, 0, error, (const char * const *)args->pdata))
     {
-      g_autofree char *commandline = flatpak_quote_argv ((const char **)args->pdata);
-      g_prefix_error (error, "running `%s`: ", commandline);
```
alexlarsson: Why remove the commandline from the error message? Seems useful for debugging.

src/builder-manifest.c
```diff
@@ -3802,12 +3786,57 @@ builder_manifest_install_dep (BuilderManifest *self,
...
+
+  ref = flatpak_build_untyped_ref (runtime,
+                                   version,
+                                   builder_context_get_arch (context));
+
+  commit = flatpak_info (opt_user, opt_installation, "--show-commit", ref, NULL);
+
+  if (commit != NULL)
+    {
+      if (builder_manifest_install_single_dep(ref, NULL, opt_user, opt_installation,
```
alexlarsson: Missing space before parentesis.

src/builder-manifest.c
```diff
@@ -3802,12 +3786,57 @@ builder_manifest_install_dep (BuilderManifest *self,
...
+                                              error))
+      {
+        return TRUE;
+      }
+    }
+  else
+    {
+      for(const char *remote = *remotes; remote != NULL; remote = *(++remotes))
```
alexlarsson: missing space

src/builder-manifest.c
```diff
@@ -3802,12 +3786,57 @@ builder_manifest_install_dep (BuilderManifest *self,
...
+	{
+	  if (builder_manifest_install_single_dep(ref, remote, opt_user, opt_installation,
```
alexlarsson: missing space

src/builder-manifest.c
```diff
@@ -3802,12 +3786,57 @@ builder_manifest_install_dep (BuilderManifest *self,
...
+		         error))
```
alexlarsson: You can't pass error multiple times like this. If any of the previous calls in the loop has failed, then error is non-null, and then if *this* call fails you'll get a g_error warning about trying to set an already error:ed GError.

You need to use a new local_error here every time, and then save a copy of one of the errors to return from the loop if nothing succeeded. I suggest keeping the first error, that is traditional.

src/builder-manifest.c
nanonyme: I left it out before I could figure out what to do with multiple errors. Your suggestion below solves that.

src/builder-manifest.c
nanonyme: Hmm, not quite clear what you meant by the function being small. Do you think it is too small? My C "how small is too small" sense is still in development.

src/builder-manifest.c
nanonyme: Yes, glib ranted about this to me once already when I had accidental infinite loop. Will fix this.

src/builder-manifest.c
alexlarsson: I mean, its small enough that just duplicating it is fine instead of sharing some stuff and then have a weird if in the middle.

alexlarsson: Oh, i see your confusion. My wording "split into two" is not great. I mean, create two different functions based on that code, one with the code for the update codepath, and one with the code for the install codepath.

src/builder-manifest.c
nanonyme: I *think* I now got it to work right. Tested by passing flathub as remote multiple times when refs were only available from freedesktop-sdk test repo.

src/builder-manifest.c
nanonyme: This is now added back.

src/builder-manifest.c
nanonyme: Having separate functions does make the code nicer to write as well. Split now. As not-regular-C-coder wasn't sure if tiny functions was a discouraged thing or not.

src/builder-manifest.c
nanonyme: More or less like you described now.
```
flatpak-builder --user --install-deps-from=flathub --repo=steam --force-clean build-dir com.valvesoftware.Steam.yml
Dependency Sdk: org.freedesktop.Sdk 19.08
Installing org.freedesktop.Sdk/x86_64/19.08 from flathub
error: Nothing matches org.freedesktop.Sdk in remote flathub
Error installing deps: running `flatpak --user install flathub -y --noninteractive org.freedesktop.Sdk/x86_64/19.08`: Child process exited with code 1
```

```
flatpak-builder --user --install-deps-from flathub --install-deps-from=flathub --repo=steam --force-clean build-dir com.valvesoftware.Steam.yml
Dependency Sdk: org.freedesktop.Sdk 19.08
Trying to install org.freedesktop.Sdk/x86_64/19.08 from flathub
error: Nothing matches org.freedesktop.Sdk in remote flathub
Trying to install org.freedesktop.Sdk/x86_64/19.08 from flathub
error: Nothing matches org.freedesktop.Sdk in remote flathub
Error installing deps: running `flatpak --user install flathub -y --noninteractive org.freedesktop.Sdk/x86_64/19.08`: Child process exited with code 1
```

src/builder-manifest.c
nanonyme: This is not done. I haven't figured out how to disambiguate the flatpak running errors from other errors.

src/builder-manifest.c
nanonyme: Can't spot where the missing space was from the diff. Did I fix this while fixing other issues?

src/builder-manifest.c
nanonyme: Hmm, looks like I misread your suggestion and now even the single-remote case prints out the remote. Is this fine?

src/builder-manifest.c
nanonyme: Now checking G_SPAWN_EXIT_ERROR domain for the GError, this seems to work.

nanonyme: Hmm, the logic I wrote is actually subtly incorrect and it's now sometimes setting error even if we succeed.

nanonyme: The cleanest way I could figure out is just collecting error always but finally resetting it if the installation was successful. Is this ok?

nanonyme:
Hm, oops, sorry. I didn't know Github did that. I was expecting it to keep the removed commits easily visible like Bitbucket and Gitlab do.

src/builder-manifest.c
```diff
@@ -3777,37 +3843,44 @@ builder_manifest_install_dep (BuilderManifest *self,
...
+	  {
+	    current_error = g_steal_pointer(error);
+	    return TRUE;
+	  }
+	  else if (*error == NULL)
+	    {
+	      *error = g_steal_pointer(&current_error);
+	    }
+	  if ((*error)->domain != G_SPAWN_EXIT_ERROR)
+	    return FALSE;
```
alexlarsson: The error handing here is a bit weird, as you're basically using the error remote arg as a local store for the error on the first error and then unsetting it when something later succeeds. I would recommend instead that you store the errors only in local variables and then call g_propagate_error (error, xxx_error) the line before you return FALSE.

Like, its not going to happen in this code, but in general you should always treat GError **error out variable as potentially NULL if the caller was not interested in error details.

nanonyme: I agree the code is a bit weird and will change as you asked but if I've understood correctly, that's essentially the end result. Stealing nulls the original.

alexlarsson: Yeah, the result will be the same (if error is not NULL), but its good to have the code be more obvious.

alexlarsson:
@rh-atomic-bot r+

PR #286, opened 2019-05-14
Ensure shallow repo mirror is updated when the git ref changes

### LINKED ISSUE #285: Error when ref changes and repo mirrored with a shallow clone
Re-running flatpak-builder with a different git ref for an already mirrored, shallow repo [reports the following error](http://sdkbuilder.gnome.org/logs/build-2019-05-14-072002/build-gnome-apps-nightly-master-org.gnome.Geary-x86_64.txt):

```
========== Building org.gnome.Geary ================
error: pathspec 'mainline' did not match any file(s) known to git.
Can't check out manifest repo: Child process exited with code 1
```

This occurs because the shallow clone only contains the ref that was specified when creating the repo. The fix is straight forward, and coming up.

mjog:
Shallow clones only contain the ref that was cloned, and so if a
mirror is shallow and the ref is changed in a subsequent invocation of
builder, the new ref is absent and git fails with an error.

This patch avoids bailing out too soon when mirroring a repo if the
requested ref does not exist, to ensure it is fetched.

Fixes #285

### DIFF
--- a/src/builder-git.c
+++ b/src/builder-git.c
@@ -531,8 +531,10 @@ builder_git_mirror_repo (const char     *repo_location,
       else
         cached_git_dir = g_object_ref (cache_mirror_dir);
 
-      /* If we're not updating, only pull from cache to avoid network i/o */
-      if (!update)
+      /* If the ref already exists (it may not with a shallow mirror
+       * if it has changed) and we're not updating, only pull from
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+

PR #288, opened 2019-05-17
Feature: Use CCACHE_DIR from environment, if set

oleid:
I'm using a Jenkins CI, which builds lots of things inside a docker container. Recently, flatpak-support was added to that container, so I  started using it. Very nice experience, so far - thanks a lot! :unicorn: 

The pipeline provides a large, shared ccache directory on a fast SSD using the standard `CCACHE_DIR` environment variable, however flatpak ignores it and places the ccache inside the state directory. I know one can change the location of the state directory, but this is IMHO too coarse. The size of the build dir is often a lot larger than a ccache. Symlinking the ccache dir won't work due to bind mounts.

This MR passes the directory configured in `CCACHE_DIR` to the sandbox, if the environment variable is set.

### DIFF
--- a/src/builder-context.c
+++ b/src/builder-context.c
@@ -202,7 +202,18 @@ builder_context_constructed (GObject *object)
   self->build_dir = g_file_get_child (self->state_dir, "build");
   self->cache_dir = g_file_get_child (self->state_dir, "cache");
   self->checksums_dir = g_file_get_child (self->state_dir, "checksums");
-  self->ccache_dir = g_file_get_child (self->state_dir, "ccache");
+
+  // Check, if CCACHE_DIR is set in environment and use it, instead of subdir of state_dir
+  const char * env_ccache_dir = g_getenv ("CCACHE_DIR");
+  if (env_ccache_dir && g_path_is_absolute(env_ccache_dir))
+    {
+      g_debug ("Using CCACHE_DIR '%s'", env_ccache_dir);
+      self->ccache_dir = g_file_new_for_path (env_ccache_dir);
+    }
+    else
+    {
+      self->ccache_dir = g_file_get_child(self->state_dir, "ccache");
+    }
 }
 
 static void
--- a/doc/flatpak-builder.xml
+++ b/doc/flatpak-builder.xml
@@ -344,7 +344,8 @@
                 <term><option>--ccache</option></term>
 
                 <listitem><para>
-                     Enable use of ccache in the build (needs ccache in the sdk)
+                     Enable use of ccache in the build (needs ccache in the sdk). The default ccache folder can be
+                     overridden by setting the environment variable CCACHE_DIR.
                 </para></listitem>
             </varlistentry>
 


TingPing:
It should probably be an env var specific to `flatpak-builder` (`FLATPAK_BUILDER_CCACHE_DIR`) and it should probably retain the tree hierarchy to avoid mixing ccache from multiple sandboxes.

src/builder-context.c
```diff
@@ -202,7 +202,18 @@ builder_context_constructed (GObject *object)
   self->build_dir = g_file_get_child (self->state_dir, "build");
   self->cache_dir = g_file_get_child (self->state_dir, "cache");
   self->checksums_dir = g_file_get_child (self->state_dir, "checksums");
-  self->ccache_dir = g_file_get_child (self->state_dir, "ccache");
+
+  // Check, if CCACHE_DIR is set in environment and use it, instead of subdir of state_dir
+  const char * env_ccache_dir = g_getenv ("CCACHE_DIR");
+  if (env_ccache_dir)
```
TingPing: Probably ensure its a remotely valid path: `g_path_is_absolute()`

oleid:
> It should probably be an env var specific to `flatpak-builder` (`FLATPAK_BUILDER_CCACHE_DIR`) and it should probably retain some hierarchy to avoid mixing ccache from multiple sandboxes.

Thanks for the review! :)

I thought about an `flatpak-builder` specific env var, too. But then I came to the conclusion, that `CCACHE_DIR` would be the name, most people would expect. Since it's nothing flatpak specific about that environment variable.  Even in _Gentoo Linux's_ build system _portage_ you call the variable `CCACHE_DIR`.

People could be surprised by `flatpak-builder` suddenly supporting `CCACHE_DIR`, when it was ignored before, I thought, as it's _the_ variable you set, when compiling with `ccache`. But then I wondered: _what could go wrong?_

* People set then env variable on purpose, especially when doing so _system wide_.
* There is no harm in sharing that folder with different applications (i.e. Gentoo does it all the time for all it's packages), as the hash of the preprocessed source is used as name of the cache file. _On the contrary_, if certain builds share sources or even use the same libraries, then you get a speed up.
* If the shared cache is too small, however, then you'll get more cache misses. Yet, you shouldn't configure a tiny ccache folder system-wide anyway.

So as you can see, I think `CCACHE_DIR` should be the name, however, if you're not convinced, then `FLATPAK_BUILDER_CCACHE_DIR` would be okay for me as well.

src/builder-context.c
oleid: You're right, thanks! fixed.

TingPing:
> There is no harm in sharing that folder with different applications (i.e. Gentoo does it all the time for all it's packages), as the hash of the preprocessed source is used as name of the cache file. On the contrary, if certain builds share sources or even use the same libraries, then you get a speed up.

It is not shared by different applications it is effectively shared against completely different operating systems.

Is it still safe? Perhaps, but I'm confident a lot of variables are at play that nobody designed it for or tests it against.

oleid:
According to [ccache docs](https://ccache.dev/manual/latest.html), the compiler binary's _file size_ and _modification time_ is included in the hash by default.  Also, different compilers usually have different C/C++ header versions, which produce different preprocessed files, even for the same source code.  AFAIK the command line args of the compiler are considered as well, as well as the path.

Different machines shouldn't be a problem; the ccache docs even mentions _sharing a cache on NFS_, so I'm positive they had this use case in mind.

TingPing:
Alright.

alexlarsson:
@rh-atomic-bot r+

PR #291, opened 2019-06-11
Use flatpak_spawnv() more

mwleeds:
This means when flatpak-builder runs a flatpak command in a subprocess,
we can see the arguments passed to flatpak in the flatpak-builder
output, if -v was used.


alexlarsson:
@rh-atomic-bot r+

PR #295, opened 2019-07-20
source-git: Add option to disable submodules

lnjX:


JBBgameich:
This is useful to prevent unnecessary issues introduced by trying to check out submodules even if they are not needed. Example: https://binary-factory.kde.org/job/Kaidan_flatpak/77/console

mwleeds:
Looks like this needs to be documented in `doc/flatpak-manifest.xml`

lnjX:
@mwleeds Done :)

alexlarsson:
@rh-atomic-bot r+

PR #303, opened 2019-08-29
doc/flatpak-builder: Clarify --run permissions

mwleeds:
Looking at the implementation of builder_manifest_run(), it seems the
permissions used for flatpak-builder --run are the same as for the final
app, with the exception of filesystem permissions. So add that caveat to
the man page.


alexlarsson:
@rh-atomic-bot r+

PR #307, opened 2019-09-13
Support --show-manifest

alexlarsson:
We want to use this in flathub to avoid all the problems we've having
parsing json manifests via the python parser, which isn't *quite* compatible
with json-glib (differs in comment and multiline string support for instance).

### DIFF
--- a/src/builder-main.c
+++ b/src/builder-main.c
@@ -47,6 +47,7 @@ static gboolean opt_build_only;
 static gboolean opt_finish_only;
 static gboolean opt_export_only;
 static gboolean opt_show_deps;
+static gboolean opt_show_manifest;
 static gboolean opt_disable_download;
 static gboolean opt_disable_updates;
 static gboolean opt_ccache;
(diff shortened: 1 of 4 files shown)


src/builder-main.c
```diff
@@ -156,6 +158,12 @@ static GOptionEntry show_deps_entries[] = {
   { NULL }
 };
 
+static GOptionEntry show_manifest_entries[] = {
+  { "verbose", 'v', 0, G_OPTION_ARG_NONE, &opt_verbose, "Print debug information during command processing", NULL },
+  { "show-manifest", 0, 0, G_OPTION_ARG_NONE, &opt_show_manifest, "Print out the manifest file in standard json format (see --show-manifest --help)", NULL },
```
ramcq: Shouldn't you remove `(see --show-manifest --help)` as this list will only appear when invoking that command? Same above too I guess.

src/builder-main.c
```diff
@@ -673,6 +689,13 @@ main (int    argc,
       return 0;
     }
 
+  if (is_show_manifest)
+    {
+      g_autofree char *json = builder_manifest_serialize (manifest);
+      g_print ("%s", json);
```
ramcq: Worth considering `%s\n` just for command line friendliness?

alexlarsson:
@rh-atomic-bot r+

PR #308, opened 2019-09-13
Support duplicate module names

alexlarsson:
If two modules happen to have the same name we uniquify them by appending "-$n". While you shouldn't normally rely on this it is sometimes hard to avoid when you're including some json snippet you don't have control over.


alexlarsson:
@rh-atomic-bot r+

PR #330, opened 2020-02-18
Support passing --disable-rofiles-fuse to --run

barthalion:


alexlarsson:
This is an extremely dangerous operation. Typically the files in /app (ie. the build dir) are hardlinks into the flatpak-builder build cache, and unless you use rofiles-fuse the container can do modifications of the files in there which changes whats in the cache breaking it in unexpected ways.

We do allow --disable-rofiles-fuse in the build stages because we then are in control of what we're running against and can ensure the cache checksouts are not hardlinks (by setting force_copy=TRUE in the cache ostree checkout operation). This makes the build slower but its safe. However, "flatpak-builder --run" just uses whatever is currently in the builddir, and there is no guarantee that what is there was built with --disable-rofiles-fuse.

Maybe we can store some kind of hint in the builddir when it contains a hardlinked cache checkout and allow disabling rofiles if that is not there. In fact, if we did that we could automatically skip the rofiles fuse when it is not necessary.

What exactly is your usecase for this?

barthalion:
There are GNOME pipelines using `flatpak-builder --run builddir` to retain SDK extensions and run meson and ninja. I don't know the full rationale why the build isn't self-contained in the first place though. Other pipelines also run number of `flatpak build` commands on builddir created by flatpak-builder but I haven't tested them yet if they work in the Docker contaner.

The use case is that GNOME doesn't have ephemeral builders anymore so I'm trying to strip CI builds from `--privileged`.

barthalion:
These two, to be exact, which are later re-used by most projects.

https://gitlab.gnome.org/GNOME/citemplates/blob/master/flatpak/flatpak-ci-initiative-sdk-extensions.yml
https://gitlab.gnome.org/GNOME/citemplates/blob/master/flatpak/flatpak_ci_initiative.yml

alexlarsson:
Ok then. I think any time we do builder_cache_checkout () with options.force_copy = FALSE, we should also create a file `.hardlinked-cache` in the checkout dir, and we should then filter this out in `commit_filter()` in the same file. Then flatpak-builder run should be changed to only enable rofuse if this file exists in the builddir.

PR #331, opened 2020-02-26
Support passing --token-type to build-export

dbnicholson:
Provide a --token-type command line option and a token-type manifest
property that allows passing the token type to build-export. The
manifest property takes precendence over the CLI option. In either case,
the value will be passed to build-export if it's >= 0. This requires
flatpak >= 1.6 to use the --token-type option in build-export.

### DIFF
--- a/doc/flatpak-builder.xml
+++ b/doc/flatpak-builder.xml
@@ -429,6 +429,19 @@
             </varlistentry>
 
             <varlistentry>
+                <term><option>--token-type=VAL</option></term>
+
+                <listitem><para>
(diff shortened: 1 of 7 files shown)


dbnicholson:
I added a fixup that makes the 32 bit value handling a bit more explicit.

dbnicholson:
> I added a fixup that makes the 32 bit value handling a bit more explicit.

I see the CI doesn't like fixups. I intend to squash this but wanted to leave it here to show the delta from what I started with.

PR #355, opened 2020-07-01
Fix ordering in "flatpak install" command

mwleeds:
Order options before positional arguments when installing dependencies.
This makes no functional difference as far as I know but it looks
cleaner.


mwleeds:
I did verify that `--install-deps-from` still works with this patch.

PR #359, opened 2020-07-22
Ignore the `--metadata` argument of flatpak build-finish

ssssam:
An app may use `--metadata` in the `finish-args` section to
add section/key/value entries in the metadata file.

If we pass this through to `flatpak build`, it's interpreted to
be a filename. Since it's not a filename, this leads to the
metadata file not being found, and an error is raised like this:

    /home/sam/src/gnome-photos/.flatpak-builder/rofiles/rofiles-qtBvVQ not initialized, use flatpak build-init

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -4135,7 +4135,8 @@ builder_manifest_run (BuilderManifest *self,
               !g_str_has_prefix (arg, "--runtime") &&
               !g_str_has_prefix (arg, "--command") &&
               !g_str_has_prefix (arg, "--extra-data") &&
-              !g_str_has_prefix (arg, "--require-version"))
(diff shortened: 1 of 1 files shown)


TingPing:
This is a fair workaround but maybe `flatpak build` should support a way pass through metadata rules?

ssssam:
> This is a fair workaround but maybe `flatpak build` should support a way pass through metadata rules?

The metadata is correctly written to the `metadata` file during `flatpak-builder` execution, as far as I can see. This is then picked up when we do `flatpak-builder --run`.

TingPing:
Ah duh. Good debuggin :)

ssssam:
Thanks for the quick review! :+1:

Issue #348, opened 2020-05-16
--run unclear


vinipsmaker:
The documentation states (emphasis mine):

>Run a command in a sandbox based on the build dir. [...] and the same permissions the final app will have (**except filesystem permissions**)

What exactly are the exceptions? Has every app that I've run through `--run` been executed outside of a sandbox? It's unclear to me what are the exceptions and I can't find this info anywhere.

TingPing:
Its certainly in a sandbox.

The exact line that comment is talking about is here: https://github.com/flatpak/flatpak-builder/blob/aec06b39452539fb894c4b222c980763642066e8/src/builder-manifest.c#L4123-L4141

vinipsmaker:
> Its certainly in a sandbox.

If so, how did the application managed to write to `.local/share/Bisq` w/o a matching `--persist` in my `"finish-args"`?

TingPing:
Sandbox is a broad term. It is in a sandbox, its permissions just aren't what you expect.

vinipsmaker:
> The exact line that comment is talking about is here:

I don't understand how this piece of code relates to give full-access to my home dir (making the whole point of sandbox moot). You sure this is the line? Care to explain how permissions leak from my specification out of that?

cxrvh:
I think the missing sandbox is a bug. You can work around it by adding `--nofilesystem=host`.

PR #362, opened 2020-09-25
--run commands in a sandbox

cxrvh:
Closes: #348


swick:
I could certainly remove the special casing in `flatpak build`. The question really is if it breaks things.

bbhtt:
I think we would want this PR too along with the Flatpak change. Latest Flatpak builder can be run against an old version of Flatpak.

PR #382, opened 2021-03-20
Port to meson

Vanadiae:
So I started a meson port, which compiles and runs fine. But before I proceed to finish completely the port I have a few questions first.

1. Would a meson port be acceptable?
2. Should I drop the autotools buildsystem to avoid having them deviate in quality as time passes?
3. The main reason I want to port flatpak-builder to meson is because I'd like to make flatpak-builder available too as a library that GNOME Builder could use to avoid all the duplication that happens between both codebases. That *could* be done with autotools, but I really don't want nor know how to write autoconf stuff. That would allow GNOME Builder to do a more fine grained treatment on the modules and sources of a manifest, like updating modules A, B, and E but not *all* modules, as we've been willing to do in https://gitlab.gnome.org/GNOME/gnome-builder/-/merge_requests/355, without having to expand flatpak-builder's CLI for every single feature Builder would need. There wouldn't be any need for a very strict API stability since only a narrow set of projects  would actually use this library (though I don't see any reason for breaking the current API which is already very nice IMO).

If `3.` is not something acceptable then I won't finish the port since well… my interest in porting to meson is in using flatpak-builder as library in GNOME Builder.

-------

Depends on https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/23 getting fixed/merged (in the meantime deleting the content of `glnx-missing-syscall.h` makes it build alright).

There's a few FIXMEs or TODOs in the code~~but in general the biggest things left to port are tests/ and doc/.~~ Done.

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -1,0 +1,63 @@
+xsltproc_args = [
+  '--nonet',
+  '--stringparam', 'man.output.quietly', '1',
+  '--stringparam', 'funcsynopsis.style', 'ansi',
+  '--stringparam', 'man.th.extra1.suppress', '1',
+  '--stringparam', 'man.authors.section.enabled', '0',
+  '--stringparam', 'man.copyright.section.enabled', '0',
+]
+xsltproc = find_program('xsltproc')
+xmlto = find_program('xmlto')
+
+# TODO: is the docbook DTD/XSL catalog check needed or does xsltproc take care of that?
+configured_xml = configure_file(
+  configuration: {
+    'VERSION': meson.project_version(),
+    # FIXME: that's not a good way of this it…
+    'srcdir': meson.current_source_dir(),
+  },
+  input: 'flatpak-builder-docs.xml.in',
+  # depend_files:
+  output: 'flatpak-builder-docs.xml',
+)
+
+man_gen_command = [
+  xsltproc, xsltproc_args,
+  'http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl',
+]
+
+man_pages = [
+  ['flatpak-builder', '1'],
+  ['flatpak-manifest', '5'],
+]
+foreach man_page : man_pages
+  man_page_name = '@0@.@1@'.format(man_page[0], man_page[1])
+  custom_target(man_page[0] + '-man',
+    command: [man_gen_command, man_page_name],
+    depend_files: man_page[0] + '.xml',
+    output: man_page_name,
+    install: true,
+    install_dir: get_option('mandir') / 'man' + man_page[1],
+  )
+endforeach
+
+doc_dir = get_option('datadir') / 'doc' / meson.project_name()
+html = custom_target('docbook',
+  command: [
+    xmlto,
+    '--skip-validation',
+    'xhtml-nochunks',
+    '-o', meson.current_build_dir(),
+    '-m', files('xmlto-config.xsl'),
+    '@INPUT@',
+  ],
+  input: configured_xml,
+  output: 'flatpak-builder-docs.html',
+  install: true,
+  install_dir: doc_dir,
+)
+
+install_data(
+  files('docbook.css'),
+  install_dir: doc_dir,
+)
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -1,63 +1,65 @@
-xsltproc_args = [
-  '--nonet',
-  '--stringparam', 'man.output.quietly', '1',
-  '--stringparam', 'funcsynopsis.style', 'ansi',
-  '--stringparam', 'man.th.extra1.suppress', '1',
-  '--stringparam', 'man.authors.section.enabled', '0',
-  '--stringparam', 'man.copyright.section.enabled', '0',
-]
-xsltproc = find_program('xsltproc')
-xmlto = find_program('xmlto')
+xsltproc = find_program('xsltproc', required: get_option('docs'))
+xmlto = find_program('xmlto', required: get_option('docs'))
 
-# TODO: is the docbook DTD/XSL catalog check needed or does xsltproc take care of that?
-configured_xml = configure_file(
-  configuration: {
-    'VERSION': meson.project_version(),
-    # FIXME: that's not a good way of this it…
-    'srcdir': meson.current_source_dir(),
-  },
-  input: 'flatpak-builder-docs.xml.in',
-  # depend_files:
-  output: 'flatpak-builder-docs.xml',
-)
+if xmlto.found() and xsltproc.found()
 
-man_gen_command = [
-  xsltproc, xsltproc_args,
-  'http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl',
-]
-
-man_pages = [
-  ['flatpak-builder', '1'],
-  ['flatpak-manifest', '5'],
-]
-foreach man_page : man_pages
-  man_page_name = '@0@.@1@'.format(man_page[0], man_page[1])
-  custom_target(man_page[0] + '-man',
-    command: [man_gen_command, man_page_name],
-    depend_files: man_page[0] + '.xml',
-    output: man_page_name,
-    install: true,
-    install_dir: get_option('mandir') / 'man' + man_page[1],
+  # TODO: is the docbook DTD/XSL catalog check needed or does xsltproc take care of that?
+  configured_xml = configure_file(
+    configuration: {
+      'VERSION': meson.project_version(),
+      # FIXME: that's not a good way of this it…
+      'srcdir': meson.current_source_dir(),
+    },
+    input: 'flatpak-builder-docs.xml.in',
+    # depend_files:
+    output: 'flatpak-builder-docs.xml',
   )
-endforeach
 
-doc_dir = get_option('datadir') / 'doc' / meson.project_name()
-html = custom_target('docbook',
-  command: [
-    xmlto,
-    '--skip-validation',
-    'xhtml-nochunks',
-    '-o', meson.current_build_dir(),
-    '-m', files('xmlto-config.xsl'),
-    '@INPUT@',
-  ],
-  input: configured_xml,
-  output: 'flatpak-builder-docs.html',
-  install: true,
-  install_dir: doc_dir,
-)
+  man_pages = [
+    ['flatpak-builder', '1'],
+    ['flatpak-manifest', '5'],
+  ]
+  foreach man_page : man_pages
+    man_page_name = '@0@.@1@'.format(man_page[0], man_page[1])
+    custom_target(man_page[0] + '-man',
+      command: [
+        xsltproc,
+        '--nonet',
+        '--stringparam', 'man.output.quietly', '1',
+        '--stringparam', 'funcsynopsis.style', 'ansi',
+        '--stringparam', 'man.th.extra1.suppress', '1',
+        '--stringparam', 'man.authors.section.enabled', '0',
+        '--stringparam', 'man.copyright.section.enabled', '0',
+        '--output', '@OUTPUT@',
+        'http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl',
+        '@INPUT@'
+      ],
+      input: man_page[0] + '.xml',
+      output: man_page_name,
+      install: true,
+      install_dir: get_option('mandir') / 'man' + man_page[1],
+    )
+  endforeach
 
-install_data(
-  files('docbook.css'),
-  install_dir: doc_dir,
-)
+  doc_dir = get_option('datadir') / 'doc' / meson.project_name()
+  html = custom_target('docbook',
+    command: [
+      xmlto,
+      '--skip-validation',
+      'xhtml-nochunks',
+      '-o', meson.current_build_dir(),
+      '-m', files('xmlto-config.xsl'),
+      '@INPUT@',
+    ],
+    input: configured_xml,
+    output: 'flatpak-builder-docs.html',
+    install: true,
+    install_dir: doc_dir,
+  )
+
+  install_data(
+    files('docbook.css'),
+    install_dir: doc_dir,
+  )
+
+endif
--- a/meson.build
+++ b/meson.build
@@ -41,9 +41,7 @@ endforeach
 debugedit = find_program('debugedit')
 
 subdir('src')
-if get_option('docs')
-  subdir('doc')
-endif
+subdir('doc')
 if get_option('tests')
   subdir('tests')
 endif
--- a/meson.build
+++ b/meson.build
@@ -38,7 +38,7 @@ foreach arg : project_c_args
 endforeach
 
 # The debugedit program is a hard dependency
-debugedit = find_program('debugedit')
+debugedit = find_program('debugedit', version: '>= 5.0')
 
 subdir('src')
 subdir('doc')
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,49 @@
+project(
+  'flatpak-builder',
+  'c',
+  license: 'LGPL-2.1-or-later',
+  meson_version: '>= 0.56.2',
+  version: '1.3.1',
+  default_options: 'c_std=gnu99',
+)
+
+cc = meson.get_compiler('c')
+
+project_c_args = [
+  '-Werror=empty-body',
+  '-Werror=strict-prototypes',
+  '-Werror=missing-prototypes',
+  '-Werror=implicit-function-declaration',
+  '-Werror=format=2',
+  '-Werror=format-security',
+  '-Werror=format-nonliteral',
+  '-Werror=pointer-arith',
+  '-Werror=init-self',
+  '-Werror=missing-declarations',
+  '-Werror=return-type',
+  '-Werror=overflow',
+  '-Werror=int-conversion',
+  '-Werror=parentheses',
+  '-Werror=incompatible-pointer-types',
+  '-Werror=misleading-indentation',
+  '-Werror=missing-include-dirs',
+  # Necessary for non POSIX stuff in glibc like fallocate
+  '-D_GNU_SOURCE',
+]
+
+foreach arg : project_c_args
+  if cc.has_argument(arg)
+    add_project_arguments(arg, language: 'c')
+  endif
+endforeach
+
+# The debugedit program is a hard dependency
+debugedit = find_program('debugedit')
+
+subdir('src')
+if get_option('docs')
+  subdir('doc')
+endif
+if get_option('tests')
+  subdir('tests')
+endif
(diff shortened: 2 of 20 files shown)


mwleeds:
For reference the discussion on porting Flatpak to meson was here: https://github.com/flatpak/flatpak/issues/2241

However perhaps it's fine for flatpak-builder to be ported to meson before Flatpak.

>Should I drop the autotools buildsystem to avoid having them deviate in quality as time passes?

Yes I think so, or if there's a reason to keep it add a CI check to ensure it's maintained.

Vanadiae:
> For reference the discussion on porting Flatpak to meson was here: [flatpak/flatpak#2241](https://github.com/flatpak/flatpak/issues/2241)
> However perhaps it's fine for flatpak-builder to be ported to meson before Flatpak.
> 

There's also https://github.com/flatpak/flatpak/pull/4141 related to this topic. If there's a need to support e.g. rhel7, then I guess the meson requirement could be lowered by not using newer features. But then, albeit running flatpak apps is important even for older systems, building flatpak apps on such systems seems a bit less important and common.
 
> > Should I drop the autotools buildsystem to avoid having them deviate in quality as time passes?
> 
> Yes I think so, or if there's a reason to keep it add a CI check to ensure it's maintained.

Yeah, I don't think it makes sense to keep both I agree.

Thanks for commenting! :)

subprojects/libglnx.wrap
```diff
@@ -0,0 +1,3 @@
+[wrap-git]
+url = https://gitlab.gnome.org/GNOME/libglnx.git
```
smcv, line 2: I think libglnx needs to be a git submodule like it is in the Autotools build system, or a [git subtree](https://github.com/git/git/blob/master/contrib/subtree/git-subtree.txt) like it is in [steam-runtime-tools](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/). This can be achieved by moving the pre-existing submodule into `subprojects/` and adjusting the Autotools build to cope with that change, like I did in <https://github.com/flatpak/flatpak/pull/4141>. That change can be made as a separate PR that this one would depend on.

Three reasons for this:

* `libglnx` is not a shared library and does not have API or ABI backwards-compatibility guarantees, so we should "pin" it to a known-good version for this version of flatpak-builder instead of always using `HEAD`.
* If we're doing a bug-fix-only release of flatpak-builder (for a serious bug or a security vulnerability) and we want LTS distributions to update their flatpak-builder to the new release, then we won't want that release to include unnecessary updates to `libglnx HEAD` that make the change harder to review.
* Meson automatically includes a submodule or subtree when doing a `meson dist`, even without `--include-subprojects`, preserving the Autotools behaviour that a known-good `libglnx` is bundled in our source releases. We need this because `libglnx` is not designed to be installed system-wide. `meson dist --include-subprojects` would add a copy of `libglnx`, but that option is undesirable because if we start including `subprojects/*.wrap` for projects that *are* designed to normally be installed system-wide, like Flatpak or GLib, we will *not* want to bundle convenience copies of *those* in source releases.

All of these considerations apply more strongly to Flatpak itself than to flatpak-builder, but I think it makes sense for a Meson port of flatpak-builder to follow the same style that we'll want to use in Flatpak.

meson_options.txt
```diff
@@ -0,0 +1,22 @@
...
+  type : 'string',
+  value : '',
+  description : 'The directory where the dwarf.h header is located, if not in include directories.',
+)
+option('documentation',
+  type : 'boolean',
+  value : false,
+  description : 'Whether to build the documentation'
+)
+option('docbook-docs',
```
smcv, line 16: AIUI, Meson style is to use `words_with_underscores` for options. If in doubt, copy the option name from a GNOME project.

meson_options.txt
```diff
@@ -0,0 +1,22 @@
...
+  type : 'feature',
+  value: 'enabled',
+  description : 'Whether to support YAML manifests',
+)
+option('dwarf_header_dir',
+  type : 'string',
+  value : '',
+  description : 'The directory where the dwarf.h header is located, if not in include directories.',
+)
+option('documentation',
```
smcv, line 11: Please use whatever the corresponding option is called in GNOME/GLib projects, to maximize the ability to share an option value between projects. If this is controlling the gtk-doc-generated reference documentation, I think that's usually `gtk_doc`?

meson.build
```diff
@@ -0,0 +1,110 @@
...
+inc_dirs = include_directories(
+  '.',
+# FIXME: handle $prefixes/libdwarf/ too or let it to the option? The later would be easier
+  # Dwarf header stuff
+  get_option('dwarf_header_dir'),
+)
+
+cc = meson.get_compiler('c')
+
+# FIXME: this was in the autoconf file but it didn't prevent from configuring. Also, it seems like fdwalk isn't called in the code?!
```
smcv, line 43: This stuff in `configure.ac` might be left over from when flatpak-builder was part of Flatpak.

meson.build
```diff
@@ -0,0 +1,110 @@
...
+
+if not cc.has_header('dwarf.h', include_directories: inc_dirs)
+  error('dwarf.h header could not be found in either the include directories or @0@'
+    .format(get_option('dwarf_header_dir'))
+  )
+endif
+
+libglnx_dep = subproject('libglnx').get_variable('libglnx_dep')
+libyaml_dep = dependency('yaml-0.1', required: get_option('yaml'))
+libelf_dep = dependency('libelf', version: '>=0.8.12')
```
smcv, line 60: If I remember correctly, libelf has not always had pkg-config metadata - but hopefully all Linux distros are now using `elfutils`, which does have it. See a recent GLib version's build system, specifically `gio/meson.build`.

GLib's `gio/meson.build` suggests that the fallback to `cc.find_library` and `cc.has_function` is only necessary on \*BSD. If that's true, then we don't (need to) care about that case for Flatpak, because Flatpak is Linux-specific and is not portable to \*BSD anyway.

subprojects/libglnx.wrap
smcv: > This can be achieved by moving the pre-existing submodule into subprojects/ and adjusting the Autotools build to cope with that change, like I did in flatpak/flatpak#4141. That change can be made as a separate PR that this one would depend on

#386

meson.build
smcv: I removed some of it in #387.

subprojects/libglnx.wrap
Vanadiae: > If we're doing a bug-fix-only release of flatpak-builder (for a serious bug or a security vulnerability) and we want LTS distributions to update their flatpak-builder to the new release, then we won't want that release to include unnecessary updates to libglnx HEAD that make the change harder to review.
> Meson automatically includes a submodule or subtree when doing a meson dist, even without --include-subprojects, preserving the Autotools behaviour that a known-good libglnx is bundled in our source releases. We need this because libglnx is not designed to be installed system-wide. meson dist --include-subprojects would add a copy of libglnx, but that option is undesirable because if we start including subprojects/*.wrap for projects that are designed to normally be installed system-wide, like Flatpak or GLib, we will not want to bundle convenience copies of those in source releases.

Oh I didn't though about that aspect. I used a wrap file because it was more convenient as meson would fetch the submodule automatically when building for the first time, but now I see the point of having it as submodule then.

meson.build
Vanadiae: Yeah, I'll drop the corresponding meson parts that are dropped in your #387 MR then.

meson.build
Vanadiae: Nice, I was surprised this manual header check was needed, so I'll remove it :)

meson_options.txt
Vanadiae: But it's not gtk-doc, it's docbook/xmlto stuff. But anyway it seems like this option is duplicated with docbook-docs, since I just copied the available autoconf options but it only enabled checks for DTD catalogs and XSL stylesheets, which should be done when enabling the main docbook docs option. So I'll just drop it.

meson_options.txt
Vanadiae: Will change that.

Vanadiae:
So I just pushed a new revision, that includes support for building the docbook documentation and manual pages, and the tests. I also pushed the changes for the resolved threads. I didn't fix the last opened thread to wait until the other MR is merged to rebase on top of it.

Some few things left to do later are to drop the autotools buildsystem, and update the CI file to use meson.

@smcv: WDYT of `3.` in my bullet list in the first message? Does that sounds reasonable to you?

Vanadiae:
Also, there's some hacks and things that doesn't quite work yet but I've added FIXMEs for input.

smcv:
> WDYT of 3. in my bullet list in the first message? Does that sounds reasonable to you?

Sorry, I don't think I'm the right person to say whether providing a library API from flatpak-builder is desirable.

subprojects/libglnx.wrap
eli-schwartz: You can do both. Have a .wrap file which points to the same url as the git submodule. When meson tries to configure the subproject, it will check to see if you're building from a git clone and the subproject is a valid submodule -- if so, it will run `git submodule update --init subprojects/foobar` for you.

This then also works if you're not building from git, but from a tarball that does NOT have the subproject properly disted. meson will simply treat it like a normal wrap-git, and clone it for you.

doc/meson.build
```diff
@@ -0,0 +1,63 @@
...
+]
+foreach man_page : man_pages
+  man_page_name = '@0@.@1@'.format(man_page[0], man_page[1])
+  custom_target(man_page[0] + '-man',
+    command: [man_gen_command, man_page_name],
+    depend_files: man_page[0] + '.xml',
+    output: man_page_name,
+    install: false,
+  )
+  install_man(man_page_name)
```
eli-schwartz, line 41: Unfortunately, this will not actually work since install_man expects to find files in the source directory at configure time. You can just set the custom target to install: true and `install_dir: join_paths(get_option('mandir'), 'man'+man_page[1])`

nanonyme:
> So I just pushed a new revision, that includes support for building the docbook documentation and manual pages, and the tests. I also pushed the changes for the resolved threads. I didn't fix the last opened thread to wait until the other MR is merged to rebase on top of it.
> 
> Some few things left to do later are to drop the autotools buildsystem, and update the CI file to use meson.
> 
> @smcv: WDYT of `3.` in my bullet list in the first message? Does that sounds reasonable to you?

I don't think this is a good idea. That implies there being a stable API. It would probably be better than some other project would provide this API and flatpak-builder and GNOME Builder would both consume it. flatpak the project is one well-maintained projec with maintenance branches which would be a good obvious pick for hosting a common API for building flapaks.

GeorgesStavracas:
I'm picking up this branch, will polish it a little, and push the improvements here. Hope you're all okay with this.

Vanadiae:
Yes I'm fine with that. Thanks for picking it up :) I personally went as far as motivation goes, so it's nice to see someone picking it up and bringing a more bareable buildsystem to flatpak-builder.

meson.build
```diff
@@ -0,0 +1,67 @@
...
+  license: 'LGPL-2.1-or-later',
+  meson_version: '>= 0.56.2',
+  version: '1.3.1',
+  default_options: 'c_std=gnu99',
+)
+
+cc = meson.get_compiler('c')
+
+add_project_arguments(
+  '-Werror=empty-body',
```
tristan957, line 13: I would also maybe wrap this in `cc.get_supported_arguments()`. Makes it a little easier to support clang.

meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option('yaml', type: 'feature', value: 'enabled', description: 'Whether to support YAML manifests')
+option('dwarf_header_dir', type: 'string', value: '',
+       description: 'The directory where the dwarf.h header is located, if not in include directories.')
+option('docs', type: 'boolean', value: false, description: 'Whether to build the DocBook documentation and man pages.')
+option('tests', type: 'boolean', value: true, description: 'Whether to run the unit tests')
```
tristan957, line 5: Whether to build and enable the unit tests?

meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option('yaml', type: 'feature', value: 'enabled', description: 'Whether to support YAML manifests')
+option('dwarf_header_dir', type: 'string', value: '',
+       description: 'The directory where the dwarf.h header is located, if not in include directories.')
+option('docs', type: 'boolean', value: false, description: 'Whether to build the DocBook documentation and man pages.')
```
tristan957, line 4: Inconsistent period usage in description

src/builder-post-process.c
```diff
@@ -28,7 +28,7 @@
 #include <sys/statfs.h>
 
 #include <gio/gio.h>
-#include "libglnx/libglnx.h"
+#include "libglnx.h"
```
tristan957, line 31: Why did the include path change? Not familiar with this lib

src/meson.build
```diff
@@ -0,0 +1,50 @@
...
+glib_req = '>= 2.44'
+flatpak_builder_deps = [
+  dependency('glib-2.0', version: glib_req),
+  dependency('gio-2.0', version: glib_req),
+  dependency('gio-unix-2.0', version: glib_req),
+  dependency('json-glib-1.0'),
+  dependency('libcurl'),
+  dependency('libelf', version: '>= 0.8.12'),
+  dependency('libglnx', default_options: ['tests=false']),
+  dependency('libsoup-2.4'),
```
tristan957, line 37: Might be worth adding a subproject for libsoup

src/meson.build
```diff
@@ -0,0 +1,50 @@
...
+  dependency('libxml-2.0', version: '>= 2.4'),
+  dependency('ostree-1', version: '>= 2017.14'),
+  dependency('yaml-0.1', required: get_option('yaml')),
```
tristan957, line 40: I could point you toward autotools wrap I did for libyaml if you want to easily support a subproject for it.

subprojects/libglnx.wrap
```diff
@@ -0,0 +1,6 @@
...
+revision = head
+
+[provide]
+libglnx = libglnx_dep
```
tristan957, line 6: Does libglnx do a meson.override_dependency()?

tests/meson.build
```diff
@@ -0,0 +1,21 @@
+test_env = environment()
+test_env.set('FLATPAK_TESTS_DEBUG', '1')
+test_env.prepend('PATH', meson.current_build_dir() / '..' / 'src', separator: ':')
+
+test_names = [
+  'test-builder',
+  'test-builder-deprecated',
+  'test-builder-python'
```
tristan957, line 8: trailing comma?

meson.build
```diff
@@ -0,0 +1,67 @@
...
+# FIXME: handle $prefixes/libdwarf/ too or let it to the option? The later would be easier
+  # Dwarf header
+  dwarf_header_dir,
+)
+
+if not cc.has_header('dwarf.h', include_directories: inc_dirs)
+  error('dwarf.h header could not be found in either the include directories or @0@'.format(dwarf_header_dir))
+endif
+
+debugedit = find_program('debugedit')
```
tristan957, line 48: What is the function of this? You don't seem to do anything with it.

meson.build
```diff
@@ -0,0 +1,67 @@
...
+dwarf_header_dir = get_option('dwarf_header_dir')
+
+inc_dirs = include_directories(
+  '.',
+# FIXME: handle $prefixes/libdwarf/ too or let it to the option? The later would be easier
+  # Dwarf header
+  dwarf_header_dir,
+)
+
+if not cc.has_header('dwarf.h', include_directories: inc_dirs)
```
tristan957, line 44: Can you do a `cc.find_library()` for this?

meson.build
```diff
@@ -0,0 +1,67 @@
+project(
+  'flatpak-builder',
+  'c',
+  license: 'LGPL-2.1-or-later',
+  meson_version: '>= 0.56.2',
+  version: '1.3.1',
```
tristan957, line 6: I might recommend using a `files('VERSION')` paradigm here. Makes it easy to share a version between autotools and Meson I would think

eli-schwartz:
> Tests are technically in TAP format, but Meson's TAP parser is unhappy with the test scripts, even though they're fine. Don't set the 'tap' protocol for tests for now (but still run them).

As witnessed in this CI run: https://github.com/flatpak/flatpak-builder/actions/runs/3342041605/jobs/5533854367

I took a look at this. The error in question was removed in Meson 0.63.0 via commit https://github.com/mesonbuild/meson/commit/a7e458effadbc884eacf34528df3a57b60e43fe3

However, the error is really just saying your TAP output stream is NOT in fact in a valid format. You have unknown result lines that cannot be parsed as TAP:
- parsers must not treat it as a failing test, but ignore it or convey it as a distinct status to the harness
- some harnesses then ignore it, some warn about it, some error about it. Some changed their approach.

Output from e.g. programs being run to implement the test, is *supposed* to be wrapped in `# ` prefixes to indicate to TAP that they are tests.

I think Meson should just warn about it, not ignore it. So I submitted a PR: https://github.com/mesonbuild/meson/pull/10964

Note that historically it was an error, for one release it was a silent pass, and now I want it to be a pass, but message you to tell you not to do that. For portability reasons and to avoid only supporting Meson >=0.63.0 you want to use properly compliant TAP anyway...

src/builder-post-process.c
GeorgesStavracas: Because this is the correct path. Meson adds strictly the correct include path here, whereas Autotools does that AND adds a catch-all include on subprojects. I'm not going to change this on Autotools now.

src/meson.build
GeorgesStavracas: ... this, on the other hand, sounds like a good idea!

meson.build
GeorgesStavracas: Just forcing the build system to fail if debugedit is not present. Unlike the Autotools build, system-installed debugedit is a hard dependency for Meson.

meson.build
GeorgesStavracas: Given that the Autotools build is going to be removed in a not too distant future, I'm happy not to intertwine both in any meaningful way.

src/meson.build
GeorgesStavracas: Added

meson.build
GeorgesStavracas: Done

meson.build
GeorgesStavracas: After taking another look at this, I realized that all this code to find dwarf.h came from building `flatpak-builder-debugedit`. The Meson build doesn't do that - it hard depends on a system debugedit. This entire part of the build system can be removed! So can the `dwarf_header_dir` option.

nanonyme: Is it really feasible already to assume system to have dwarf5 capable debugedit?

GeorgesStavracas: It's reasonable to assume that anything equal or newer than Ubuntu 22.04 has that, which what the Meson build requires at minimum.

meson.build
tristan957: Right, but just because something exists on the build system doesn't mean it is available on the target system. I would remove this check and let flatpak-builder fail gracefully if the tool isn't found. This is the discussion we previously had in #mesonbuild

meson.build
nanonyme: Fair, looks like separate debugedit project has become common since when this feature was originally implemented.

PR #386, opened 2021-04-06
Move libglnx submodule into subprojects/

smcv:
* autogen.sh: Remove remnants of bundled copy of bubblewrap

* Move libglnx submodule into subprojects/
    
    This is a step towards being able to build using Meson rather than
    Autotools (#382). Meson has built-in support for subprojects, but requires
    them to appear in the top-level subprojects/ directory.
    
    We have to add -Isubprojects because we include <libglnx/libglnx.h>.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "libglnx"]
-	path = libglnx
+	path = subprojects/libglnx
 	url = https://gitlab.gnome.org/GNOME/libglnx.git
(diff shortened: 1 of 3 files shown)


smcv:
> We have to add -Isubprojects because we include <libglnx/libglnx.h>

An alternative way to fix this would be to change all our uses of libglnx to `#include <libglnx.h>`, which seems to be the canonical way to use it, as seen in #382 - but that can be done later, if desired.

mwleeds (approved):

PR #387, opened 2021-04-06
Clean up dead code in build system

smcv:
* autogen.sh: Remove remnants of bundled copy of bubblewrap

* build: Remove obsolete check for fdwalk
    
    Maybe Flatpak calls this, but flatpak-builder doesn't.

* build: Remove obsolete check for sys/capability.h
    
    Nothing in flatpak-builder or libglnx needs this.

* build: Remove instructions for maintaining libtool interface versioning
    
    flatpak-builder doesn't have a shared library.

* build: Comment why we check for <sys/xattr.h>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,18 +1,5 @@
 AC_PREREQ([2.63])
 
-# Making releases:
-#   FLATPAK_MICRO_VERSION += 1;
-#   FLATPAK_INTERFACE_AGE += 1;
-#   FLATPAK_BINARY_AGE += 1;
-# if any functions have been added, set FLATPAK_INTERFACE_AGE to 0.
-# if backwards compatibility has been broken,
-# set FLATPAK_BINARY_AGE and FLATPAK_INTERFACE_AGE to 0.
-#
-# in easier to understand terms:
-#
-# on the stable branch, interface age == micro
-# on the unstable (ie master), interface age = 0
-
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -91,9 +78,8 @@ AX_VALGRIND_CHECK
 
 PKG_PROG_PKG_CONFIG([0.24])
 
-AC_CHECK_FUNCS(fdwalk)
+# For libglnx
 AC_CHECK_HEADER([sys/xattr.h], [], [AC_MSG_ERROR([You must have sys/xattr.h from glibc])])
```
Vanadiae, line 82: Since it's for libglnx, I wonder why is this check not in libglnx itself?

smcv: Mainly because libglnx doesn't have a complete build system, only a fragment of one.

I actually did start to move this into `libglnx.m4`, but then I looked at libglnx's `meson.build` and noticed that it had never actually checked for this...

In practice, "your glibc doesn't have `sys/xattr.h`" is probably synonymous with "your glibc is from the distant past", so maybe we can just lose this check? The oldest glibc I have convenient access to is eglibc 2.15 from Ubuntu 12.04 (2012), which does have it. Apparently musl also has it.

Vanadiae: If a system as old as 2012 has it, then it's probably fairly safe to drop the check entirely.

PR #403, opened 2021-07-18
Add inline data source

gasinvein:
In some cases the user may need to embed small files into the manifest. This is especially common practice for automatic manifest generators, namely flatpak-builder-tools. Currently, inline data can be stored as a `file` type source with `data:` URL.

This adds a specialized source type that writes given text content to a file (optionally, non-text data can be base64-encoded).
It provides an alternative to `data:` URLs, with following advantages:
- Less overhead from character escaping, thus, smaller manifest size. For yaml, overhead is reduced to none. For json, only a few special characters needs to be escaped.
- Much easier to write by hand (and read by a human), especially for yaml. For example, an embedded `.yarnrc` can be turned from this
  ```yaml
  type: file
  url: data:yarn-offline-mirror%20/run/build/vscode/flatpak-node/yarn-mirror%0A--install.offline%20true%0A--run.offline%20true%0A
  dest: main
  dest-filename: .yarnrc
  ```
  to this
  ```yaml
  type: inline
  contents: |
    yarn-offline-mirror /run/build/vscode/flatpak-node/yarn-mirror
    --install.offline true
    --run.offline true
  dest: main
  dest-filename: .yarnrc
  ```

### DIFF
--- a/src/builder-source-inline.c
+++ b/src/builder-source-inline.c
@@ -1,0 +1,233 @@
+#include "config.h"
+
+#include <string.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/statfs.h>
+
+#include "builder-flatpak-utils.h"
+#include "builder-utils.h"
+#include "builder-source-inline.h"
+
+struct BuilderSourceInline
+{
+  BuilderSource parent;
+
+  char         *contents;
+  gboolean      base64;
+  char         *dest_filename;
+};
+
+typedef struct
+{
+  BuilderSourceClass parent_class;
+} BuilderSourceInlineClass;
+
+G_DEFINE_TYPE (BuilderSourceInline, builder_source_inline, BUILDER_TYPE_SOURCE);
+
+enum {
+  PROP_0,
+  PROP_CONTENTS,
+  PROP_BASE64,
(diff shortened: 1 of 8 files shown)


src/builder-source-inline.c
```diff
@@ -0,0 +1,233 @@
...
+
+static gboolean
+builder_source_inline_validate (BuilderSource  *source,
+                                GError        **error)
+{
+  BuilderSourceInline *self = BUILDER_SOURCE_INLINE (source);
+
+  if (self->dest_filename != NULL &&
+      strchr (self->dest_filename, '/') != NULL)
+    return flatpak_fail (error, "No slashes allowed in dest-filename");
```
TingPing, line 112:
```suggestion
    return flatpak_fail (error, "No slashes allowed in dest-filename, use dest property for directory");
```

TingPing:
Looks and sounds reasonable to me.

src/builder-source-inline.c
gasinvein: This was copied over from script source. Maybe we should change this message for all the other source types?

TingPing: I think so, its just helpful.

gasinvein: Done.

PR #406, opened 2021-07-24
Add 'no-resolved-manifest' option

GeorgesStavracas:
Some applications have security tokens passed by their build system. In these cases, having the resolved manifest file distributed with the sandbox is a security issue.

Add an option to not create the resolved manifest file.

### DIFF
--- a/src/builder-context.c
+++ b/src/builder-context.c
@@ -81,6 +81,7 @@ struct BuilderContext
   gboolean        have_rofiles;
   gboolean        run_tests;
   gboolean        no_shallow_clone;
+  gboolean        no_resolved_manifest;
 
   BuilderSdkConfig *sdk_config;
 };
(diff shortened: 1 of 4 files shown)


GeorgesStavracas:
I really don't know if this is enough to prevent the creation of `/app/manifest.json`...

TingPing:
I think this is a bit heavy handed disabling the entire manifest.

Maybe we could just have a `secrets-env` property that isn't copied to the manifest but everything else remains?

refi64:
Would some sort of "secret sources" have a slightly lower risk of ending up embedded inside? (I'd *hope* random build tasks aren't embedding the entire environment, but I have unfortunately seen that before...)

GeorgesStavracas:
> Maybe we could just have a `secrets-env` property that isn't copied to the manifest but everything else remains?

Hm, could you please expand this idea? Specifically, how would it be propagated to the compile flags of a specific module?

TingPing:
> Hm, could you please expand this idea? Specifically, how would it be propagated to the compile flags of a specific module?

Well its an env var, so `--foo=$BAR`.

Obviously you still need to have secure build infra that doesn't leak build logs and such.

PR #409, opened 2021-08-16
Use debugedit to parse debug info for DWARF 5 support

refi64:
fd.o SDK 21.08 brings with it GCC 11, which enables the use of DWARF 5
for debug info by default. Unfortunately, flatpak-builder's
DWARF-parsing code was pulled from a rather old version of rpm's
debugedit, meaning it does not have DWARF 5 support. This new DWARF
version *does* bring quite a few new operand forms, so updating the
current code to support verison 5 would likely be nontrivial.

In the time since flatpak-builder first pulled rpm's debugedit code,
however, debugedit has become a standalone project. Thus, we can now
simply depend on the new standalone debugedit to parse DWARF files. The
code requires debugedit 5.0+ (which is the first stable release, the odd
version number is just so it doesn't conflict with rpm versions), and it
will fall back to building its own debugedit in the same manner as
Flatpak can build its own bwrap/xdg-dbus-proxy.


refi64:
Well, the CI image doesn't have a new-enough libdw to be able to handle DWARF5, which doesn't bode well for some of the systems flatpak-builder still needs to run on... In particular, libdw >= 1.171/1.172 is needed.

I'm actually not sure what the best solution would be here, since, we can't ignore DWARF5 defines on platforms where they're not available (flatpak-builder would not run properly when trying to extract debuginfo when building any flatpaks with runtime 21.08). Ubuntu 18.04 only has libdw 1.170, with no newer version available in backports. I guess maybe the PPA could add the newer libdw? But that might take a while to pan out...

src/builder-utils.c
```diff
@@ -534,998 +534,75 @@ builder_gobject_from_data (GType       gtype,
...
+  data_stream = g_data_input_stream_new (input_stream);
+  files = g_ptr_array_new_with_free_func (g_free);
 
-  for (i = 0; i < ehdr.e_shnum; ++i)
+  while (TRUE)
     {
-      scns[i] = elf_getscn (elf, i);
-      gelf_getshdr (scns[i], &shdr[i]);
-    }
+      ;
```
alexlarsson, line 575: Whats with this semicolon?

src/builder-utils.c
```diff
@@ -534,998 +534,75 @@ builder_gobject_from_data (GType       gtype,
...
 
-  elf_flagelf (elf, ELF_C_SET, ELF_F_LAYOUT);
-
-  shdr = g_new0 (GElf_Shdr, ehdr.e_shnum);
-  scns =  g_new0 (Elf_Scn *, ehdr.e_shnum);
+  input_stream = g_unix_input_stream_new (tmp_fd, FALSE);
+  data_stream = g_data_input_stream_new (input_stream);
+  files = g_ptr_array_new_with_free_func (g_free);
 
-  for (i = 0; i < ehdr.e_shnum; ++i)
```
alexlarsson, line 1455: At this point you should unlink the tempfile so that it will be gone once tmp_fd is closed. Otherwise we will leak temp files on disk.

alexlarsson:
I would be fine with not updating flatpak-builder on 18.04. For backwards compat, the main important part is the user facing part, not the developer side. The flathub builders are ubuntu 20.04 anyway.

@smcv @mwleeds opinions?

refi64:
@alexlarsson issues resolved (including the mysterious, lonesome semicolon), worth noting that @smcv did agree to dropping 18.04 support on the Matrix chat.

alexlarsson:
Poor poor semicolon...

PR #418, opened 2021-09-21
WIP: Port to libappstream

pwithnall:
appstream-glib is mostly unmaintained, and appstream is more
actively developed (and up to date with the AppStream specification).

This ports flatpak-builder to use the command line tools from appstream
(`appstreamcli` and `appstreamcli-compose`) rather than the ones from
appstream-glib (`appstream-util`, `appstream-builder` and
`appstream-compose`).

This should result in no changes for most things, but should fix
propagation of `<requires>`/`<recommends>` elements, screenshot
`<caption>`s, and other elements which have been added to the AppStream
specification recently.

For example, see https://github.com/flathub/flathub/issues/2439

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -2403,16 +2403,16 @@ static GFile *
 builder_manifest_find_appdata_file (BuilderManifest *self,
                                     GFile           *app_root)
 {
-  /* We order these so that share/appdata/XXX.appdata.xml if found
+  /* We order these so that share/metainfo/$FLATPAK_ID.metainfo.xml is found
      first, as this is the target name, and apps may have both, which will
      cause issues with the rename. */
   const char *extensions[] = {
-    ".appdata.xml",
     ".metainfo.xml",
+    ".appdata.xml",
   };
   const char *dirs[] = {
-    "share/appdata",
     "share/metainfo",
+    "share/appdata",
   };
   g_autoptr(GFile) source = NULL;
 
@@ -2511,16 +2511,15 @@ builder_manifest_cleanup (BuilderManifest *self,
       appdata_source = builder_manifest_find_appdata_file (self, app_root);
       if (appdata_source)
         {
-          /* We always use the old name / dir, in case the runtime has older appdata tools */
-          g_autoptr(GFile) appdata_dir = g_file_resolve_relative_path (app_root, "share/appdata");
-          g_autofree char *appdata_basename = g_strdup_printf ("%s.appdata.xml", self->id);
+          g_autoptr(GFile) appdata_dir = g_file_resolve_relative_path (app_root, "share/metainfo");
+          g_autofree char *appdata_basename = g_strdup_printf ("%s.metainfo.xml", self->id);
 
           appdata_file = g_file_get_child (appdata_dir, appdata_basename);
 
           if (!g_file_equal (appdata_source, appdata_file))
             {
               g_autofree char *src_basename = g_file_get_basename (appdata_source);
-              g_print ("Renaming %s to share/appdata/%s\n", src_basename, appdata_basename);
+              g_print ("Renaming %s to share/metainfo/%s\n", src_basename, appdata_basename);
 
               if (!flatpak_mkdir_p (appdata_dir, NULL, error))
                 return FALSE;
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -2292,9 +2292,9 @@ cmpstringp (const void *p1, const void *p2)
 }
 
 static gboolean
-appstream_compose (GFile   *app_dir,
-                   GError **error,
-                   ...)
+appstreamcli_compose (GFile   *app_dir,
+                      GError **error,
+                      ...)
 {
   g_autoptr(GPtrArray) args = NULL;
   const gchar *arg;
@@ -2306,7 +2306,7 @@ appstream_compose (GFile   *app_dir,
   g_ptr_array_add (args, g_strdup ("--die-with-parent"));
   g_ptr_array_add (args, g_strdup ("--nofilesystem=host:reset"));
   g_ptr_array_add (args, g_file_get_path (app_dir));
-  g_ptr_array_add (args, g_strdup ("appstream-compose"));
+  g_ptr_array_add (args, g_strdup ("appstreamcli-compose"));
 
   va_start (ap, error);
   while ((arg = va_arg (ap, const gchar *)))
@@ -2316,7 +2316,7 @@ appstream_compose (GFile   *app_dir,
 
   if (!builder_maybe_host_spawnv (NULL, NULL, 0, error, (const char * const *)args->pdata, NULL))
     {
-      g_prefix_error (error, "ERROR: appstream-compose failed: ");
+      g_prefix_error (error, "ERROR: appstreamcli-compose failed: ");
       return FALSE;
     }
 
@@ -2751,14 +2751,18 @@ builder_manifest_cleanup (BuilderManifest *self,
 
       if (self->appstream_compose && appdata_file != NULL)
         {
+          g_autofree char *app_root_path = g_file_get_path (app_root);
+          g_autofree char *prefix_arg = g_strdup_printf ("--prefix=%s", app_root_path);
           g_autofree char *basename_arg = g_strdup_printf ("--basename=%s", self->id);
-          g_print ("Running appstream-compose\n");
-          if (!appstream_compose (app_dir, error,
-                                  self->build_runtime ?  "--prefix=/usr" : "--prefix=/app",
-                                  "--origin=flatpak",
-                                  basename_arg,
-                                  self->id,
-                                  NULL))
+          g_autofree char *components_arg = g_strdup_printf ("--components=%s", self->id);
+          g_print ("Running appstreamcli-compose\n");
+          if (!appstreamcli_compose (app_dir, error,
+                                     self->build_runtime ?  "--prefix=/usr" : "--prefix=/app",
+                                     "--origin=flatpak",
+                                     "--result-root=/",
+                                     components_arg,
+                                     "/",
+                                     NULL))
             return FALSE;
         }
 
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -2292,8 +2292,7 @@ cmpstringp (const void *p1, const void *p2)
 }
 
 static gboolean
-appstreamcli_compose (GFile   *app_dir,
-                      GError **error,
+appstreamcli_compose (GError **error,
                       ...)
 {
   g_autoptr(GPtrArray) args = NULL;
@@ -2301,12 +2300,8 @@ appstreamcli_compose (GFile   *app_dir,
   va_list ap;
 
   args = g_ptr_array_new_with_free_func (g_free);
-  g_ptr_array_add (args, g_strdup ("flatpak"));
-  g_ptr_array_add (args, g_strdup ("build"));
-  g_ptr_array_add (args, g_strdup ("--die-with-parent"));
-  g_ptr_array_add (args, g_strdup ("--nofilesystem=host:reset"));
-  g_ptr_array_add (args, g_file_get_path (app_dir));
-  g_ptr_array_add (args, g_strdup ("appstreamcli-compose"));
+  g_ptr_array_add (args, g_strdup ("appstreamcli"));
+  g_ptr_array_add (args, g_strdup ("compose"));
 
   va_start (ap, error);
   while ((arg = va_arg (ap, const gchar *)))
@@ -2314,9 +2309,9 @@ appstreamcli_compose (GFile   *app_dir,
   g_ptr_array_add (args, NULL);
   va_end (ap);
 
-  if (!builder_maybe_host_spawnv (NULL, NULL, 0, error, (const char * const *)args->pdata, NULL))
+  if (!flatpak_spawnv (NULL, NULL, 0, error, (const char * const *)args->pdata, NULL))
     {
-      g_prefix_error (error, "ERROR: appstreamcli-compose failed: ");
+      g_prefix_error (error, "ERROR: appstreamcli compose failed: ");
       return FALSE;
     }
 
@@ -2751,17 +2746,17 @@ builder_manifest_cleanup (BuilderManifest *self,
 
       if (self->appstream_compose && appdata_file != NULL)
         {
-          g_autofree char *app_root_path = g_file_get_path (app_root);
-          g_autofree char *prefix_arg = g_strdup_printf ("--prefix=%s", app_root_path);
-          g_autofree char *basename_arg = g_strdup_printf ("--basename=%s", self->id);
           g_autofree char *components_arg = g_strdup_printf ("--components=%s", self->id);
-          g_print ("Running appstreamcli-compose\n");
-          if (!appstreamcli_compose (app_dir, error,
-                                     self->build_runtime ?  "--prefix=/usr" : "--prefix=/app",
+          g_autofree char *app_root_path = g_file_get_path (app_root);
+          g_autofree char *result_root_arg = g_strdup_printf ("--result-root=%s", app_root_path);
+
+          g_print ("Running appstreamcli compose\n");
+          if (!appstreamcli_compose (error,
+                                     "--prefix=/",
                                      "--origin=flatpak",
-                                     "--result-root=/",
+                                     result_root_arg,
                                      components_arg,
-                                     "/",
+                                     app_root_path,
                                      NULL))
             return FALSE;
         }
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -23,7 +23,8 @@ jobs:
         libjson-glib-dev shared-mime-info desktop-file-utils libpolkit-agent-1-dev libpolkit-gobject-1-dev \
         libseccomp-dev libsoup2.4-dev libsystemd-dev libxml2-utils libgpgme11-dev gobject-introspection \
         libgirepository1.0-dev libappstream-dev libdconf-dev clang socat flatpak \
-        libcurl4-gnutls-dev libflatpak-dev libyaml-dev elfutils git patch unzip
+        libcurl4-gnutls-dev libflatpak-dev libyaml-dev elfutils git patch unzip \
+        appstream
     - name: Check out flatpak
       uses: actions/checkout@v1
       with:
@@ -54,7 +55,8 @@ jobs:
         libjson-glib-dev shared-mime-info desktop-file-utils libpolkit-agent-1-dev libpolkit-gobject-1-dev \
         libseccomp-dev libsoup2.4-dev libsystemd-dev libxml2-utils libgpgme11-dev gobject-introspection \
         libgirepository1.0-dev libappstream-dev libdconf-dev clang flatpak \
-        libcurl4-gnutls-dev libflatpak-dev libyaml-dev elfutils git patch unzip
+        libcurl4-gnutls-dev libflatpak-dev libyaml-dev elfutils git patch unzip \
+        appstream
     - name: Check out flatpak
       uses: actions/checkout@v1
       with:
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -22,7 +22,7 @@ jobs:
         libfuse-dev ostree libostree-dev libarchive-dev libcap-dev libattr1-dev libdw-dev libelf-dev \
         libjson-glib-dev shared-mime-info desktop-file-utils libpolkit-agent-1-dev libpolkit-gobject-1-dev \
         libseccomp-dev libsoup2.4-dev libsystemd-dev libxml2-utils libgpgme11-dev gobject-introspection \
-        libgirepository1.0-dev libappstream-glib-dev libdconf-dev clang socat flatpak \
+        libgirepository1.0-dev libappstream-dev libdconf-dev clang socat flatpak \
         libcurl4-gnutls-dev libflatpak-dev libyaml-dev elfutils git patch unzip
     - name: Check out flatpak
       uses: actions/checkout@v1
@@ -53,7 +53,7 @@ jobs:
         libfuse-dev ostree libostree-dev libarchive-dev libcap-dev libattr1-dev libdw-dev libelf-dev \
         libjson-glib-dev shared-mime-info desktop-file-utils libpolkit-agent-1-dev libpolkit-gobject-1-dev \
         libseccomp-dev libsoup2.4-dev libsystemd-dev libxml2-utils libgpgme11-dev gobject-introspection \
-        libgirepository1.0-dev libappstream-glib-dev libdconf-dev clang flatpak \
+        libgirepository1.0-dev libappstream-dev libdconf-dev clang flatpak \
         libcurl4-gnutls-dev libflatpak-dev libyaml-dev elfutils git patch unzip
     - name: Check out flatpak
       uses: actions/checkout@v1
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -3,10 +3,10 @@ name: Flatpak-builder CI
 on:
   push:
     branches:
-    - master
+    - main
   pull_request:
     branches:
-    - master
+    - main
 
 jobs:
   check:
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -8,7 +8,7 @@ dn=$(dirname $0)
 
 pkg_install sudo which attr fuse \
     libubsan libasan libtsan elfutils-libelf-devel libdwarf-devel \
-    elfutils git gettext-devel libappstream-glib-devel bison \
+    elfutils git gettext-devel libappstream-devel bison \
     libcurl-devel dconf-devel fuse-devel \
     /usr/bin/{update-mime-database,update-desktop-database,gtk-update-icon-cache}
 pkg_install_testing ostree-devel ostree libyaml-devel
--- a/configure.ac
+++ b/configure.ac
@@ -15,6 +15,7 @@ AC_INIT([flatpak-builder],
 GLIB_REQS=2.44
 OSTREE_REQS=2017.14
 FLATPAK_REQS=0.99.1
+APPSTREAMCLI_REQS=0.15.0
 SYSTEM_DEBUGEDIT_REQS=5.0
 LIBDW_REQS=0.172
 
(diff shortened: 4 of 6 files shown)


pwithnall:
See https://github.com/flatpak/flatpak/pull/4426 for the related flatpak changes.

CC @ximion

TingPing:
My main worry is if there are subtle differences between the behavior of the CLI tools that affect Flathub.

pwithnall:
Yeah, this is going to need some good testing.

nanonyme:
Note I don't think this is going to work. flatpak-builder is *supposed* to support newer and older runtimes so the command in sandbox probably needs to remain parameter by parameter (and command name) identical. There could probably be a shim that passes to libappstream.

ximion:
Screenshot captions? Wow, those have been part of AppStream since the February 2014 release! I had no idea this was missing!
`appstreamcli-compose` is rather new and currently not installed in $PATH to avoid issues - if it is installed, it can be run via `appstreamcli compose`, but I'm not sure yet whether I like this behavior.

Compose will cache screenshots in newer versions soon, I am not sure of you want this (feature can be disabled). There probably are differences, `appstreamcli-compose` and `appstream-compose` are quite different in what they can do, but the resulting data should be compatible with appstream-glib and older libappstream versions (older versions will just ignore any new elements). So I don't expect much breakage in other tools which just consume the metadata.

nanonyme:
As said, the very fact the command name is different is a starting point for pain here. And it looks like parameters are not. If this is merged and released, it means this flatpak-builder version can no longer be used to build *any* apps using old runtimes. This does not seem acceptable. We need some approach that will be backwards and forwards compatible.

ximion:
I see a few options we have:
* Make sure all old runtimes include AppStream, so it'd have to be retroactively added for them to work again
* Make people not use a newer flatpak-builder with older runtimes (which basically means force everyone to upgrade)
* Add some compatibility code so flatpak-builder can use appstreamcli if that's available, but fall back to appstream-util if it's not there for older runtimes

I think the last option would probably be the most compatible and the one associated with the least amount of pain.

nanonyme:
I'm pretty sure Flathub infra would not work with the second option (@barthalion can probably say something about that) so that's probably a no-go. The first option is nigh impossible. Even if we could sort out freedesktop-sdk , there are gnome runtime which need to be dealt with separately. And then there's external child runtimes.

TingPing:
The third option is the only realistic one. Supporting both.

barthalion:
I think the third option is the way to go. Ironically, I think using appstream-util from the runtime was not the best idea, so from Flathub's perspective, it would be vastly preferred to use appstream *from the host*, and fall back to appstream-util *inside the sandbox*.

ramcq:
Boo host dependency. If flatpak-builder is going to invoke a binary from outside the build environment, shouldn't it be a Flatpak? Then it's independent of whatever runtime it needs, the Flathub buildbot/flat-manager can use the same flatpak, and we can expect/assume the same behaviour from flatpak-builder of this version or later, regardless of the host or the SDK being used. (We could potentially even remove the fallback behaviour and just say, flatpak-builder after X.Y will always need org.whatever.appstream-badger to be installed.)

nanonyme:
@ramcq you mean like org.flatpak.Builder?

ramcq:
No I mean like we provide an appstreamcli equiv of `https://github.com/flathub/org.freedesktop.appstream-glib` and flatpak-builder uses that going forward. Maybe with a host fallback? But appstream is an evolving spec and we've already suffered on our infra from being pinned to a host version of appstream-glib, which is why that flatpak exists. Let's not re-invent that problem here?

barthalion:
Being a host dependency is exactly what lets us to easily decouple it from runtimes, like we already do with `org.freedesktop.appstream-glib`, which is a symlink to `appstream-util` on the build infra. I don't think tying flatpak-builder directly to a flatpak hosted on Flathub is something everyone will be happy about.

barthalion:
Host dependency means org.flatpak.Builder could be theoretically self-contained as well.

ramcq:
> Being a host dependency is exactly what lets us to easily decouple it from runtimes, like we already do with `org.freedesktop.appstream-glib`, which is a symlink to `appstream-util` on the build infra. I don't think tying flatpak-builder directly to a flatpak hosted on Flathub is something everyone will be happy about.

Alright, makes sense. Host vs flatpak can be a deployment decision.

ximion:
Since AppStream may add new things in future, having its revision pinned to the runtime is probably unexpected - when it's a host dependency, Flatpak could say "require at least version X or higher" and be done with it, while when it's part of the runtime users of older runtimes can't get new stuff until AppStream is updated in the runtime as well.
If it would be part of the host, you'd also only need to test AppStream-related changes once with one runtime, instead of verifying them with all older versions as well. I see arguments for both, but having it as host dependency does make some sense to me.

We also haven't had any breaking changes in AppStream for many years now, and the 1.0 release will just be removal of deprecated stuff (which hopefully will not break much stuff), so I think the only issue you might run into is a stricter validator (which I don't think is a huge issue, as only errors and warnings fail validation, and those are all quite severe issues that need to be fixed).

JakobDev:
What's the current state of this? Flatpak itself has already switched to libappstream in the latest pre-release and I really want to use the new features like <recommends> in my Programs.

barthalion:
Rebased it on top of the main branch and added a commit to run appstreamcli commands on the host. CI will likely break but it's a start.

barthalion:
I've kicked off a test build at https://github.com/flathub/org.flatpak.Builder/pull/81. When I manage to build it, I will do some testing on Flathub apps.

barthalion:
OK, so this is something that produces a supposedly working `appstreamcli` invocation:

```
Running appstreamcli-compose
FB: Running: appstreamcli compose --prefix= --origin=flatpak --result-root=/var/home/bpiotrowski/dev/flathub/org.gabmus.whatip/.flatpak-builder/rofiles/rofiles-q1EnZD/files --components=org.gabmus.whatip /var/home/bpiotrowski/dev/flathub/org.gabmus.whatip/.flatpak-builder/rofiles/rofiles-q1EnZD/files
Only accepting component: org.gabmus.whatip
Processing directory: /var/home/bpiotrowski/dev/flathub/org.gabmus.whatip/.flatpak-builder/rofiles/rofiles-q1EnZD/files
Composing metadata...
Success!
```

But the resulting `builddir/files/share/swcatalog/xml/flatpak.xml.gz` is empty, and I don't see `app-info` anywhere. Have I missed something about how the new compose command works?

barthalion:
`build-update-repo` confirms that it's not really correct:

```
> flatpak build-update-repo repo
Updating appstream branch
No appstream data for app/org.gabmus.whatip/x86_64/localtest: No such file or directory: /files/share/app-info
```

So I guess that `build-update-repo` needs to start reading files from the `swcatalog` directory and the compose command needs to actually produce a proper XML file.

ximion:
@barthalion The `appstreamcli compose` command will generate mostly legacy-free output - but nothing is stopping you from just adding a `mv /files/share/swcatalog /files/share/app-info` after the compose to keep things running without many changes :-)
As for the empty output file: Can you run compose with `--verbose --print-report=short` to see what's going on here? I can't reproduce this here... (the next version of AppStream will emit an error hint in case filters are set but no output was created though, to make an issue like this easily noticeable without having to check the generated output).

nanonyme:
@ximion probably makes sense to add WIP commit those flags into this branch?

barthalion:
@ximion Added `--verbose --print-report=short`, an output from failed attempt to build GoLand can be seen here: https://github.com/flathub/org.flatpak.Builder/runs/6452841114?check_suite_focus=true#step%3A10%3A2366= I'm waiting for `bot, build` to pass to poke around locally.

barthalion:
How to reproduce:

```
flatpak remote-add --user --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
flatpak install --user https://dl.flathub.org/build-repo/89476/org.flatpak.Builder.flatpakref
git clone https://github.com/flathub/com.jetbrains.GoLand
cd com.jetbrains.GoLand
flatpak run org.flatpak.Builder --verbose --user --sandbox --force-clean --repo=repo --install-deps-from=flathub builddir com.jetbrains.GoLand.yaml
```

```
Running appstreamcli-compose
FB: Running: appstreamcli compose --prefix= --origin=flatpak --verbose --print-report=short --result-root=/home/bpiotrowski/.cache/flatpak-builder/rofiles/rofiles-guPqDs/files --components=com.jetbrains.GoLand /home/bpiotrowski/.cache/flatpak-builder/rofiles/rofiles-guPqDs/files
** (appstreamcli:272): DEBUG: 15:57:11.287: run appstreamcli: compose
Only accepting component: com.jetbrains.GoLand
Processing directory: /home/bpiotrowski/.cache/flatpak-builder/rofiles/rofiles-guPqDs/files
Composing metadata...
** (appstreamcli-compose:272): DEBUG: 15:57:11.380: No media export directory set, but icon 128x128@1 is set for remote delivery. Disabling remote for this icon type.
** (appstreamcli-compose:272): DEBUG: 15:57:11.380: No media export directory set, but icon 128x128@2 is set for remote delivery. Disabling remote for this icon type.
** (appstreamcli-compose:272): DEBUG: 15:57:11.380: No media export directory set, but screenshots are supposed to be stored. Disabling screenshot storage.
** (appstreamcli-compose:272): DEBUG: 15:57:11.395: Will use temporary directory '/tmp/as-compose_5slX1E' and delete it after this run.
** (appstreamcli-compose:272): DEBUG: 15:57:11.396: Creating contents index for directory: /home/bpiotrowski/.cache/flatpak-builder/rofiles/rofiles-guPqDs/files
** (appstreamcli-compose:272): DEBUG: 15:57:11.398: Index done for directory: /home/bpiotrowski/.cache/flatpak-builder/rofiles/rofiles-guPqDs/files
(appstreamcli-compose:272): GLib-GIO-DEBUG: 15:57:11.408: _g_io_module_get_default: Found default implementation local (GLocalVfs) for ‘gio-vfs’
** (appstreamcli-compose:272): DEBUG: 15:57:11.409: Removing temporary directory '/tmp/as-compose_5slX1E'
Run failed, some data was ignored.
Errors were raised during this compose run:
general
  E: filters-but-no-output
Refer to the generated issue report data for details on the individual problems.
Error: ERROR: appstreamcli-compose failed: Child process exited with code 1
```

ximion:
@barthalion This was very helpful, thanks! The problem was that AppStream didn't expect that someone would set `--prefix=` to an empty string (which would mean "look in the current directory" which makes no sense to the code as it can't jump beyond the given root directory). Instead, it expects `/` there.
I adjusted AppStream's code to sanitize that value and work with an empty prefix as probably most people would expect, and also squashed another issue which made it count results incorrectly that was introduced in the last version.
So, taking that patch should solve your issues (and I expect a new 0.15.4 AppStream release soonish as well, likely next week).

barthalion:
@pwithnall @mwleeds @smcv Can you please review everything I pushed here since April? I will squash commits a bit once I resolve your comments, if you have any.

barthalion:
FYI @smcv, we're going to need the yet-nonexistent appstream 0.15.4 in flatpak PPA.

src/builder-manifest.c
```diff
@@ -2747,12 +2747,16 @@ builder_manifest_cleanup (BuilderManifest *self,
       if (self->appstream_compose && appdata_file != NULL)
         {
           g_autofree char *components_arg = g_strdup_printf ("--components=%s", self->id);
+          g_autofree char *app_root_path = g_file_get_path(app_root);
```
pwithnall, line 2749:
```suggestion
          g_autofree char *app_root_path = g_file_get_path (app_root);
```

src/builder-manifest.c
```diff
@@ -2403,16 +2403,16 @@ static GFile *
 builder_manifest_find_appdata_file (BuilderManifest *self,
                                     GFile           *app_root)
 {
-  /* We order these so that share/appdata/XXX.appdata.xml if found
+  /* We order these so that share/metainfo/$FLATPAK_ID.appdata.xml is found
```
pwithnall, line 2406:
```suggestion
  /* We order these so that share/metainfo/$FLATPAK_ID.metainfo.xml is found
```

pwithnall (commented):
Here’s a few things I spotted, but they’re both nitpicks. I can’t see anything conceptually wrong with this, but as evidenced by my failure to get this working earlier, I don’t really know enough of how this machinery all fits together to give a particularly useful review!

smcv:
> we're going to need the yet-nonexistent appstream 0.15.4 in flatpak PPA

Depending on a version that doesn't yet exist seems like a showstopper for merging this. We'll need an actual release.

What Ubuntu branch(es) is this needed for? 22.04? 20.04?

This is a 1.13.x feature, so it'll only be in [ppa:flatpak/development](https://launchpad.net/~flatpak/+archive/ubuntu/development).

src/builder-manifest.c
```diff
@@ -2751,14 +2745,20 @@ builder_manifest_cleanup (BuilderManifest *self,
...
+          g_autofree char *components_arg = g_strdup_printf ("--components=%s", self->id);
+          g_autofree char *app_root_path = g_file_get_path(app_root);
+          g_autofree char *result_root_arg = g_strdup_printf ("--result-root=%s", app_root_path);
+
+          g_print ("Running appstreamcli-compose\n");
+          if (!appstreamcli_compose (error,
+                                     "--prefix=",
+                                     "--origin=flatpak",
+                                     "--verbose",
+                                     "--print-report=short",
```
ximion, line 2757: You *may* want to increase verbosity for the issues found, but only if users are actually expected to read this. The `short` report will display the tag names of all the issues found, `on-error` (the default) will only display information if there was an error, and `full` will show the found issue tags and their explanatory texts for people who are less familiar with AppStream. This will increase verbosity quite a bit.
I think either `short` or `full` may be good options here.

ximion:
> Depending on a version that doesn't yet exist seems like a showstopper for merging this. We'll need an actual release.

I intend to make a 0.15.4 release either this or next week to resolve this issue.
With @smcv 's comments this PR makes sense to me as far as I understand it :-)
One thing that may be a bit annoying is that newer AppStream versions need Meson 0.62.x or higher - still easy to package, but without the unfortunate bug in 0.15.3 we may not have needed this extra step at all.
Let me know if any help is needed from my side!

ramcq:
> @barthalion This was very helpful, thanks! The problem was that AppStream didn't expect that someone would set `--prefix=` to an empty string (which would mean "look in the current directory" which makes no sense to the code as it can't jump beyond the given root directory). Instead, it expects `/` there. I adjusted AppStream's code to sanitize that value and work with an empty prefix as probably most people would expect, and also squashed another issue which made it count results incorrectly that was introduced in the last version. So, taking that patch should solve your issues (and I expect a new 0.15.4 AppStream release soonish as well, likely next week).

Is there a way to work around the behaviour in prior versions, by passing $(CWD) or . or /proc/dir or something as the prefix, so we can use more widely available versions?

smcv:
> One thing that may be a bit annoying is that newer AppStream versions need Meson 0.62.x or higher

Sorry, I am not comfortable with the amount of yak shaving implied by building the latest Meson for every branch of Ubuntu that is going to need to run the latest flatpak-builder, so that we can build appstream, so that we can build flatpak-builder. I don't even *use* Ubuntu, except as a way to test these backports from Debian...

If you want AppStream to be something that is widely backported, then being conservative with its dependencies (like Flatpak is) would seem like a good idea. You can have a policy of "always run the latest, it's easy to backport" and pay the maintenance cost of doing so, or you can have a policy of "feel free to use new dependencies to reduce the maintenance cost", but you can't have it both ways.

> just pass --prefix=/ and this should work with all available versions except 0.15.3

I believe the Flathub infra runs on Ubuntu LTS, with libappstream 0.12.0 (on 18.04), 0.12.10 (on 20.04) or 0.15.2 (on 22.04), so we might have dodged that bullet. By the time it advances to something newer than 22.04, 0.15.3 should have been superseded anyway.

If this is specifically a problem with 0.15.3, then I think appropriate pseudocode for `configure.ac` would go something like this:

* if version >= 0.15.4:
    * good, use `--prefix=` (or `--prefix=/` if that also works)
* else if version >= (some minimum) && version < 0.15.3:
    * good, use `--prefix=/`
* else if `[ -n "$APPSTREAMCLI_IGNORE_BUG_123" ]`
    * warning "appstreamcli version 0.15.3 has bug 123, but continuing since you say so..."
    * use whatever seems more likely to work
* else
    * fatal error "appstreamcli version 0.15.3 is not suitable, see bug 123"

ximion:
> Sorry, I am not comfortable with the amount of yak shaving implied by building the latest Meson for every branch of Ubuntu that is going to need to run the latest flatpak-builder, so that we can build appstream, so that we can build flatpak-builder. I don't even use Ubuntu, except as a way to test these backports from Debian...

I could help with that, if you want to.

> If you want AppStream to be something that is widely backported, then being conservative with its dependencies (like Flatpak is) would seem like a good idea.

AppStream hasn't changed its meson dependency (or GLib dependency) for half a decade before, but at some point the amount of hackery to work around broken Meson behaviour and the amount of calls to external shell scripts or Python code became a bit too much to be considered sane, especially since newer Meson versions had fixed all of the issues.

The unofficial policy for AppStream is "needs to run on the latest Debian Stable / Ubuntu LTS release", simply because I need to do backports to both to some extent in order to build metadata (in general though, dependencies are only increased if necessary, I think the last time a dependency that wasn't Meson was touched was when appstream started to depend on libxmlb which allowed bumping the GLib version as well).

Your version check logic makes sense to me :-)
> good, use --prefix= (or --prefix=/ if that also works)

`--prefix=` and `--prefix=/` yield the same behaviour on all versions, so you don't need to change this :-)

@JakobDev If this is an issue, I could quite easily revive my AppStream PPA `ppa:ximion/appstream` for Ubuntu 22.04 and provide up-to-date builds there. Shouldn't be hard to do at all, as long as PPAs are acceptable for Flathub :-)
I can possibly also provide a backport to Debian 11, if needed.

src/builder-manifest.c
barthalion: I don't think users are expected to do anything about compose output. We should probably add a separate validation step like Flathub does as part of its CI, but I guess it's out of scope here.

barthalion:
Just to clarify, while Flathub uses Ubuntu LTS, we mostly rely on Flatpak PPA for Flatpak itself. We use flatpak-builder through flatpak at https://github.com/flathub/org.flatpak.Builder/, so we can control all dependencies. I mentioned the need for having newer AppStream in some PPA as we need CI to pass here.

barthalion:
This still misses changes to Flatpak to support both `app-info` and `swconfig`, looking into that now.

JakobDev:
Does that mean, that older Flatpak versions are no longer working, after the change to appstreamcli is done? This could be a problem. even if this gets backported to older versions.

ximion:
> Does that mean, that older Flatpak versions are no longer working, after the change to appstreamcli is done? This could be a problem. even if this gets backported to older versions.

Any version of AppStream >= 0.15.2 but not 0.15.3 should work fine. Older versions *might* work too, but 0.15.2 was there release where compose is no longer marked as experimental feature and all remaining issues for it had been fixed. Its API should also be stable from the point onward.

> Just to clarify, while Flathub uses Ubuntu LTS, we mostly rely on Flatpak PPA for Flatpak itself. We use flatpak-builder through flatpak at https://github.com/flathub/org.flatpak.Builder/, so we can control all dependencies. I mentioned the need for having newer AppStream in some PPA as we need CI to pass here.

Ah, thanks, that does make sense! I can provide the PPA if needed, but I can also just create the packages to upload to a PPA you control, if you prefer that. 0.15.4 will have no breaking changes (as usual) or additions to the specification, so it should be a safe upgrade (especially over 0.15.3 with its compose issue).

barthalion:
(I apologize for the noise, turns out I've had some stray commits in the main branch and had to clean it up after wrong rebase.)

JakobDev:
> Any version of AppStream >= 0.15.2 but not 0.15.3 should work fine. Older versions might work too, but 0.15.2 was there release where compose is no longer marked as experimental feature and all remaining issues for it had been fixed. Its API should also be stable from the point onward.

Sorry if you misunderstood me, but I was talking about the Flatpak patch that barthalion linked above, that makes Flatpak also search in swcatalog and not only in app-info.

ximion:
> Sorry if you misunderstood me, but I was talking about the Flatpak patch that barthalion linked above, that makes Flatpak also search in swcatalog and not only in app-info.

Ah, I indeed misunderstood! The actual answer here depends on how this is handled inside Flatpak, so @barthalion knows this best I think.

smcv:
> We use flatpak-builder through flatpak at https://github.com/flathub/org.flatpak.Builder/, so we can control all dependencies. I mentioned the need for having newer AppStream in some PPA as we need CI to pass here.

OK, that's less concerning.

The Flatpak PPAs are really serving two or three purposes:

* they're a convenience for Ubuntu users who want to run newer Flatpak apps but are not provided with backports by their distro
* they're used by Flathub infrastructure
* they're used to run CI for flatpak-builder

For the "convenience for Ubuntu users" use-case, I don't really want to add an appstream backport to the set of packages that we aim to support, because it feels like upgrading flatpak shouldn't apply potentially disruptive changes to OS components that depend on libappstream.

For Flathub infrastructure, obviously whatever the Flathub sysadmins are comfortable with is fine.

Similarly, for CI, whatever we're OK with build-depending on is fine.

It might make sense to have a new `ppa:flatpak/for-flathub` or `ppa/flatpak:builder-development` (or something like that) alongside [ppa:flatpak/development](https://launchpad.net/~flatpak/+archive/ubuntu/development) and [ppa:flatpak/stable](https://launchpad.net/~flatpak/+archive/ubuntu/stable), so that we can put more-disruptive changes into the new PPA without causing unexpected package upgrades for users of `flatpak/development` and `flatpak/stable`?

smcv:
> Hmm, but isn't this about backporting libappstream for flatpak-builder, not flatpak itself? Former is used by developers who probably know what they're doing whereas latter is used by end-users who might not.

Yes, but if the backport is in the same PPA that's used by end-users of Flatpak, then adding that PPA as an apt source will upgrade their libappstream, whether it's actually necessary or not.

ximion:
@smcv 
> For the "convenience for Ubuntu users" use-case, I don't really want to add an appstream backport to the set of packages that we aim to support, because it feels like upgrading flatpak shouldn't apply potentially disruptive changes to OS components that depend on libappstream.

Fair enough - AppStream isn't a Flatpak-exclusive component after all, Ubuntu uses it for other purposes as well. Many AppStream upgrades are fairly safe nowadays, issues  occur specifically when newer tags are added, like e.g. the screenshot image/video localization, which software centers initially didn't know about. This is an issue with older libappstream releases, while newer versions will abstract this away so older frontends continue to work without issues. If Flatpak apps wanted to use these, upgrading AppStream for the frontends would be needed, or these features should not be used until newer AppStream releases have percolated through the various distros and packaging solutions (the strategy I've used in the past).

In any case, most stable distros have recent enough AppStream versions now and those changes aren't that frequent. And the issue here simply falls into the "unfortunate bug" category.
That said, flatpak-builder does need quite a recent AppStream version simply because a bunch of changes to help Flatpak have only been added in recent releases, as originally (lib)appstream-compose was more a utility for generating data from distribution packaging.

The proposed PPA split makes a lot of sense to me, as it would generate the absolute smallest amount of potential friction.

I've also made the latest AppStream release available for Ubuntu 22.04 in a PPA here: https://launchpad.net/~ximion/+archive/ubuntu/appstream/+packages?field.name_filter=&field.series_filter=jammy
Also, 0.15.4 is out now and contains mostly validation and compose bugfixes and enhancements :-)

ximion:
> it's unlikely to be enabled by distros. It's still disabled by default.

Fedora, Arch, Debian and Ubuntu have it enabled by default for quite a while now. You could just depend on it and distros will enable this feature ;-) (And I think we are only missing OpenSUSE to cover 99% of Linux users)

> Also it has a quite large set of dependencies which might make it uncomfortable for distros 

Extremely unlikely. It does not have more dependencies than appstream-glib had, all of the dependencies are fairly common for systems with a GUI, and only the resulting compose components depend on the GUI stuff, so can be split out.
The main AppStream library is very light on dependencies as it may be used on systems which really do not want GUI components, but the compose library also doesn't depend on anything that KDE and GNOME don't already have by default.

> One possible approach would be biting the bullet and introducing libappstream into sdk in freedesktop-sdk 22.08.

I think that would be a good idea in general. Means projects which validate their metadata in a test via appstreamcli won't need to disable that or build libappstream when creating their Flatpak, but could just run this out of the box.

nanonyme:
> I think that would be a good idea in general.
Note it's a deceptively complex idea. If flatpak-builder calls tool from runtime that is only present in new runtimes and not older ones, that imposes a version dependency on flatpak-builder and runtimes. As time goes, older runtimes would also have obsolete versions of libappstream.

nanonyme:
> > it's unlikely to be enabled by distros. It's still disabled by default.
> 
> Fedora, Arch, Debian and Ubuntu have it enabled by default for quite a while now. You could just depend on it and distros will enable this feature ;-) (And I think we are only missing OpenSUSE to cover 99% of Linux users)

Ubuntu 20.04 LTS does not build compose for libappstream as displayed by failing CI.

ximion:
Ubuntu 22.04 LTS does have it though, so if you are on the latest LTS, things just work :-)
Can the CI be upgraded to use the latest Ubuntu LTS?

nanonyme:
@ximion you're missing the point. There are like a dozen or so desktop distros basing on Ubuntu 20.04 LTS. They will gradually rebase to 22.04 LTS in a year or so. It is not correct to say atm libappstream compose is commonly available. Of course, this might not matter if people are not supposed to use flatpak-builder but instead org.flatpak.Builder.

JakobDev:
Just a question: Flathub validate Appstream files. It currently uses appstream-util, which makes sense, because it is part of appstream-glib. But the validation is not as good as appstreamcli. There are currently a few Apps with invalid Appstream on Flathub e.g. flathub/website#379. What will happen after the chance? Will the App not be able to push update until the Appstream is fixed (which is weird for someone who does not know much about the Baclground, because the Appstream was marked as valid earlier) or did it cause more problems?

ximion:
> Will the App not be able to push update until the Appstream is fixed 

That is what I would expect to happen, yes. But I also think that's okay, it would pretty much be the same thing when `appstream-util` was updated and introduced new checks, and I think it's reasonable to expect an app maintainer to fix these issues.
To make it easier for them, Flathub could pass the `--explain` option to `appstreamcli validate` - that will generate a very verbose report, but will also make it easy to get information about the found issue.

JakobDev:
@barthalion Any news on this?

ximion:
This patch actually looks fine to me - @pwithnall, which tasks are left to do, and can I help with anything?

pwithnall:
> This patch actually looks fine to me - @pwithnall, which tasks are left to do, and can I help with anything?

Sorry, I’ve kind of checked out on this PR since people started talking about host vs runtime dependencies, which are well outside my area of expertise. (I was also away for a long time this summer.)

From my point of view, there’s nothing left WIP in the PR, but you should probably be asking @barthalion.

nanonyme:
@ximion afaik the problem was that this simply does not work as is. There's another WIP branch against flatpak https://github.com/flatpak/flatpak/pull/4904 which this work supposedly depends on.

ximion:
@nanonyme That PR looks good to me, but we could also live without it by passing `--data-dir` to `appstreamcli` and redirect data to the old locations for compatibility - either solution should work.

nanonyme:
I suggest this is switched to --data-dir so it's not dependent on flatpak changes.

barthalion:
@lunarequest will try to push this forward with my guidance. We found out it needs some restructuring to execute the compose command only once and pass screenshots mirroring when needed.

ximion:
Nice, thanks! If there's anything you need me for, just ask - it also looks like I'll be able to make it to FOSDEM this year, so I'll be very easy to reach in case anyone of you is also there.

barthalion:
Superseded by https://github.com/flatpak/flatpak-builder/pull/517

PR #419, opened 2021-09-22
Re-enable HTTP compression

TingPing:
See https://github.com/flatpak/flatpak/pull/4430 for details


alexlarsson:
This looks ok, as builder doesn't have the issues that flatpak does with gzip content encoding.

PR #434, opened 2021-11-17
Fix missing [ExtensionOf] metadata for sources

mwleeds:


alexlarsson:
The bundled sources is not an extension though. Like, it has not extension point in the app, and the app does not need, nor use the bundled sources.

mwleeds:
> The bundled sources is not an extension though. Like, it has not extension point in the app, and the app does not need, nor use the bundled sources.

Makes sense. But what behavior do we want when an app is masked and its Sources ref is installed, refrain from updating the Sources? If so we need some extra logic in `flatpak_dir_needs_update_for_commit_and_subpaths()` (https://github.com/flatpak/flatpak/issues/4585)

bbhtt:
There is indeed no need to add sources extension to metadata.

PR #441, opened 2021-12-15
Use new metainfo path

### LINKED ISSUE #252: Appstream files are moved from metainfo to appdata
## Flatpak-builder version
`flatpak-builder 1.0.1`

## Description of the problem

The correct place for those files _is_ `metainfo`, see flatpak/flatpak#30, but `flatpak-builder` moves them to `appdata`, leaving `metainfo` empty.

mwleeds:
Export to `share/metainfo` rather than `share/appdata` as discussed in https://github.com/flatpak/flatpak-builder/issues/252 and https://github.com/flatpak/flatpak/pull/4600

Fixes #252


alexlarsson:
I merged https://github.com/flatpak/flatpak/pull/4600, so apps installing to appdata will end up being exported to metainfo in the end. It also means that the build installing in metainfo will work, but only with newer (unreleased) versions of flatpak, so at the very least we should not merge this until that is widely available. In fact, I wonder it we need to merge it at all, as the end result will be the same anyway (files in metainfo).

mwleeds:
I guess one reason to merge this eventually is that we want the final name to match the source name as the comment says in `builder_manifest_find_appdata_file()`

orowith2os:
@alexlarsson can you poke at this again? It's been a year and a half since the flatpak PR was merged, and it appears that the relevant Flatpak versions are available such that it can be assumed that the metainfo files are handled properly everywhere.

AsciiWolf:
Is there any update about this? Flatpak apps on Flathub still seem to have their AppStream metadata exported into the "appdata" directory with *.appdata.xml as their suffix.

nanonyme:
Is this now obsoleted by libappstream migration?

PR #450, opened 2022-01-17
Disable filesystem access with --nofilesystem=host:reset

smcv:
Flatpak 1.12.3 and 1.10.6 resolved [CVE-2022-21682](https://github.com/flatpak/flatpak/security/advisories/GHSA-8ch7-5j3h-g4fx), a vulnerability in the overall system consisting of Flatpak and flatpak-builder, but they did so in a way that caused regressions for some use patterns, by changing the semantics of `--nofilesystem=host`.

The current plan to resolve those regressions while still resolving the vulnerability is to revert those changes in Flatpak, but instead add a new command-line option `--nofilesystem=host:reset` which undoes the effect of any `--filesystem` arguments in the app's manifest or overrides.

This is expected to need Flatpak 1.12.4 and flatpak-builder 1.2.2 (or backports of a similar change to older branches) to be released to distributions as a set.

See https://github.com/flatpak/flatpak/pull/4675, https://github.com/flatpak/flatpak/pull/4678 for the Flatpak side of this.

This PR is untested, but should work - I wanted to get a PR open so that I could give it an explanatory commit message.

---

Commit description:

**Disable filesystem access with --nofilesystem=host:reset**

This requires <https://github.com/flatpak/flatpak/pull/4678>.

In addition to counteracting an earlier --filesystem=host, in Flatpak
versions that support it, the new --nofilesystem=host:reset removes all
filesystem access that might have been inherited from the app manifest
or overrides. This prevents CVE-2022-21682, while avoiding behaviour
changes in Flatpak for non-builder use cases.

In older Flatpak versions, this option acts as --filesystem=host with an
unknown mode suffix, which is ignored (with a warning, which is harmless
but will hopefully nudge people towards upgrading Flatpak to a version
that enables CVE-2022-21682 to be avoided). flatpak-builder will still
be vulnerable to CVE-2022-21682 in this case.


alexlarsson:
This matched exactly my outstanding code which i did some testing with, except i had some additional work to support --nofilesystem=host:.reset in f-b itself, which I added to this PR.

alexlarsson:
Hmm, the CI fails due to the fatal-warnings...

src/builder-flatpak-utils.c
```diff
@@ -1773,6 +1774,12 @@ parse_filesystem_flags (const char *filesystem, FlatpakFilesystemMode *mode)
       if (mode)
         *mode = FLATPAK_FILESYSTEM_MODE_CREATE;
     }
+  else if (g_str_has_suffix (filesystem, ":reset"))
+    {
+      len -= 6;
+      if (mode)
+        *mode = FLATPAK_FILESYSTEM_MODE_RESET;
+    }
```
smcv, line 1782: This makes it impossible to represent an app that has, for example, both `--nofilesystem=host:reset` and `--filesystem=host:ro`. Is that going to be a problem?

If it is, then I think we need to use the same trick as in Flatpak itself: convert `host:reset` into `{"host-reset": NONE}` in the hash table, and back to `--nofilesystem=host:reset` when serializing.

smcv: I implemented what I suggest here in ed5e692.

smcv:
Changes in 8a8828e5...a7a3e28c:

* rebase on master
* map `--nofilesystem=host:reset` to `{host-reset: NONE}` and back

alexlarsson:
This looks good to me, but maybe we should squash the two argument handling patches together?

Issue #448, opened 2022-01-10
[Bug]: Fails to start rofiles-fuse from ostree >= 2022.1 if built against FUSE 3


smcv:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a bug that matches the one I want to file, without success.

### flatpak-builder version

1.2.1

### Flatpak version

1.12.2

### How to reproduce

1. Upgrade to `ostree` 2022.1 from Debian experimental, which was compiled against FUSE 3 (ostreedev/ostree#2376)
2. Build a Flatpak app: I used libportal's convenience script `./build-aux/build.sh`

### Expected Behavior

The Flatpak app builds

### Actual Behavior

The build fails with

```
fuse: unknown option(s): `-o splice_write,splice_move'
...
error: Build directory /home/smcv/src/debian/libportal/.flatpak-builder/rofiles/rofiles-RWlNxs not initialized, use flatpak build-init
```

### Additional Information

It looks as though the options `splice_write,splice_move`, which are used unconditionally by flatpak-builder, are not valid for FUSE 3 and only valid for FUSE 2.

Workaround: install a version of `ostree` (specifically, whichever package has the `rofiles-fuse` executable) that was built against FUSE 2.

PR #451, opened 2022-01-18
context: Make FUSE 2 optimizations opt-in

smcv:
FUSE 3 does not support the splice_write,splice_move options: those
and similar speedups are enabled by default with FUSE 3.

Distributions wanting faster FUSE 2 filesystems can enable this by
configuring --with-fuse=2, but their binaries will not work when
libostree has been recompiled against FUSE 3. This behaviour was
unconditional in older releases such as 1.2.1.

Conversely, we can add opt-in features that require FUSE 3 by
configuring --with-fuse=3 if we want to, although at the moment no such
features exist.

The new default is to be compatible with all libostree versions,
regardless of whether they were compiled for FUSE 2 or 3.

Resolves: https://github.com/flatpak/flatpak-builder/issues/448

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -197,6 +197,15 @@ AM_CONDITIONAL(DOCBOOK_DOCS_ENABLED, test x$enable_docbook_docs = xyes)
 AC_ARG_VAR([XMLTO],[Define/override the 'xmlto' location.])
 AC_ARG_VAR([XMLTO_FLAGS],[Define/override 'xmlto' options, like '--skip-validation'.])
 
+AC_ARG_WITH([fuse],
+            [AS_HELP_STRING([--with-fuse={2,3}],
+                            [Only support this major version of FUSE])],
+            [fuse_major="$withval"],
+            [fuse_major=maximize-compatibility])
+AS_IF([test $fuse_major = 2],
+      [AC_DEFINE([ASSUME_FUSE_2], [],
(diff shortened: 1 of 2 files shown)


TingPing:
> FUSE 3 does not support the splice_write,splice_move options:

Typo in commit message? FUSE 2?

smcv:
This fixes compatibility with libostree 2022.1-2 in Debian experimental, which I want to proceed with soon.

Up to you whether to include this in the flatpak-builder release with #450 (which we need for CVE-2022-21682), or whether to defer it to a subsequent release.

smcv:
> > FUSE 3 does not support the splice_write,splice_move options:
> 
> Typo in commit message? FUSE 2?

No, I mean what I said. If you compile libostree against FUSE 2, then `rofiles-fuse` accepts those CLI options, and only enables the corresponding optimizations if the appropriate CLI option was given. If you compile it against FUSE 3, then `rofiles-fuse` stops accepting those CLI options (see #448) and the corresponding optimizations are unconditionally enabled.

smcv:
> > Typo in commit message? FUSE 2?
> 
> No, I mean what I said. If you compile libostree against FUSE 2, then `rofiles-fuse` accepts those CLI options

And, yes, this is really unusual: normally upgrading something gives you access to *more* APIs.

PR #452, opened 2022-01-18
Prepare v1.2.2

smcv:


smcv:
(Please don't merge this as a non-fast-forward, it matches a tag)

smcv:
Looks like Github doesn't provide a way to do this. Please could someone with sufficient admin powers (@alexlarsson?) fetch the `1.2.2` tag and do a fast-forward push to master?

Issue #375, opened 2021-01-21
Automatically detect main branch (or default branch)


A6GibKm:
## Automatically detect the default branch of a git source

It is becoming more and more common for repositories to name their default branch as main rather than master, and flatpak builder won't find this branch (or any other default master), as only master is hardcoded.

Main could also be hardcoded, or alternatively,  there could be some libgit magic to detect the proper default branch of the repo in question.

hadess:
I was pretty certain I filed that one, but looks like I didn't.

This is how it was done for jhbuild, but the branch is already checked out when that call is made:
https://gitlab.gnome.org/GNOME/jhbuild/-/commit/77192689d33d54a43289130c5d7edf85e89f9f3d

refi64:
So this is actually a bit of a mess, and it turns out *you cannot find this reliably if you're not using `git clone`*. `refs/remotes/origin/HEAD` is only created when you using `git clone`, and it contains the value of the Git internal function `guess_remote_head`. The *only* other section of the Git code where this is used is in `git remote show`, which is a porcelain command and thus cannot be parsed reliably.

refi64:
Amendment: it looks like you *maybe* can obtain it with a mirror repo, but the mechanism for finding it is completely different.

hadess:
That's pretty much how far I got when trying it out as well. I think there there might be a flag day in the not too distant future where the default branch is switched in flatpak-builder and that's that.

nbenitez:
Fortunately, it happens that `git remote show` uses `git ls-remote` underneath to obtain its output, and `git ls-remote` is a plumbing command so it's well suited for scripts.

The remote default branch name can be obtained with:\
`git ls-remote --symref origin HEAD`

I tested it and it's working fine in JHBuild with [this MR I just filed](https://gitlab.gnome.org/GNOME/jhbuild/-/merge_requests/171).

hadess:
> The remote default branch name can be obtained with:
> `git ls-remote --symref origin HEAD`

Huh, you still need the checkout to already be made, which has been the problem from the start.

nbenitez:
Ah yes, but luckily the remote name can also be a URL, so this works:\
`git ls-remote --symref https://gitlab.gnome.org/GNOME/glib.git/ HEAD`

hadess:
> Ah yes, but luckily the remote name can also be a URL, so this works:

Ha, nice!

nbenitez:
I went ahead and submitted PR for this.

PR #455, opened 2022-01-24
ask for default branch name instead of harcoding 'master'

nbenitez:
Many projects are starting to change their default branch
name from 'master' to 'main' and similar, eg.:
https://gitlab.gnome.org/GNOME/evince/-/issues/1635

So when a project is not providing a specific branch,
instead of defaulting to 'master' we should ask the
remote repo which is their default branch name.

That can be retrieve with following command:
git ls-remote --symref origin HEAD

where 'origin' can also be a URL to the repository.

Fixes issue #375

### DIFF
--- a/src/builder-git.c
+++ b/src/builder-git.c
@@ -868,3 +868,33 @@ builder_git_checkout (const char     *repo_location,
 
   return TRUE;
 }
+
+char *
+builder_git_get_default_branch (const char *repo_location)
+{
+  g_auto(GStrv) parts = NULL;
+  g_autofree char *output = NULL;
+  g_autoptr(GError) error = NULL;
+
+  if (!git (NULL, &output, 0, &error,
+            "ls-remote", "--symref", repo_location, "HEAD", NULL))
+    return g_strdup ("master");
+
+  /* Example output:
+   * $ git ls-remote --symref http://gitlab.gnome.org/GNOME/evince.git HEAD
+   * warning: redirigiendo a https://gitlab.gnome.org/GNOME/evince.git/
+   * ref: refs/heads/main	HEAD
+   * 7a5ceb841874bef4b91282eee1025b0335a21795	HEAD
+   */
+  parts = g_strsplit (output, "\t", 1);
+  if (g_strv_length (parts) > 1 && g_str_has_prefix (parts[1], "HEAD") &&
+      g_strstr_len (parts[0], -1, "ref: "))
+    {
+      char *branch = strrchr (parts[0], '/');
+      if (branch != NULL)
+        return g_strdup (branch + 1);
+    }
+
+  g_debug ("Failed to auto-detect default branch from git output");
+  return g_strdup ("master");
+}
--- a/src/builder-source-git.c
+++ b/src/builder-source-git.c
@@ -44,6 +44,7 @@ struct BuilderSourceGit
   char         *tag;
   char         *commit;
   char         *orig_ref;
+  char         *default_branch_name;
   gboolean      disable_fsckobjects;
   gboolean      disable_shallow_clone;
   gboolean      disable_submodules;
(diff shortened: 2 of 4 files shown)


src/builder-git.c
```diff
@@ -868,3 +868,32 @@ builder_git_checkout (const char     *repo_location,
 
   return TRUE;
 }
+
+char *
+builder_git_get_default_branch (const char *repo_location)
+{
+  char *default_branch;
+  char ** parts;
+  char ** parts2;
```
TingPing, line 877:
```suggestion
  g_auto(GStrv) parts;
  g_auto(GStrv) parts2;
```

This should work I believe.

src/builder-git.c
```diff
@@ -868,3 +868,32 @@ builder_git_checkout (const char     *repo_location,
...
+  g_autoptr(GError) error = NULL;
+
+  if (!git (NULL, &output, 0, &error,
+            "ls-remote", "--symref", repo_location, "HEAD", NULL))
+    return g_strdup ("master");
+
+  parts = g_strsplit (output, "\t", 1);
+  if (g_str_has_prefix (parts[1], "HEAD") && g_strstr_len (parts[0], -1, "ref: "))
+    {
+      parts2 = g_strsplit (parts[0], "/", -1);
```
TingPing, line 888: I feel like you can avoid this allocation entirely with `strrchr()` or similar?

Not actually sure the first `g_strsplit()` is really necessary either but not a huge deal.

TingPing: This might work:

```c
char *
builder_git_get_default_branch (const char *repo_location)
{
  g_auto(GStrv) parts = NULL;
  g_autofree gchar *output = NULL;
  g_autoptr(GError) error = NULL;

  if (!git (NULL, &output, 0, &error,
            "ls-remote", "--symref", repo_location, "HEAD", NULL))
    return g_strdup ("master");

  parts = g_strsplit (output, "\t", 1);
  if (g_strcmp0 (parts[1], "HEAD") == 0 && g_str_has_prefix (parts[0], "ref: "))
    {
      char *branch = strrchr (parts[0], '/');
      if (branch != NULL)
        return g_strdup (branch + 1);
    }

  return g_strdup ("master");
}
```

nbenitez: Nice 👍🏻  the `strrchr` use, I pushed your version but without the `g_str_has_prefix (parts[0], "ref: ")` because `ref: ` is possible to not be at the start of the output when there is warning of a redirect, like example below:
```
$ git ls-remote --symref http://gitlab.gnome.org/GNOME/evince.git HEAD
warning: redirigiendo a https://gitlab.gnome.org/GNOME/evince.git/
ref: refs/heads/main	HEAD
7a5ceb841874bef4b91282eee1025b0335a21795	HEAD
```

src/builder-git.c
```diff
@@ -868,3 +868,32 @@ builder_git_checkout (const char     *repo_location,
...
+   */
+  parts = g_strsplit (output, "\t", 1);
+  if (g_strv_length (parts) > 1 && g_str_has_prefix (parts[1], "HEAD") &&
+      g_strstr_len (parts[0], -1, "ref: "))
+    {
+      char *branch = strrchr (parts[0], '/');
+      if (branch != NULL)
+        return g_strdup (branch + 1);
+    }
+
```
TingPing, line 897: Hmm, since this may be somewhat fragile parsing the output. Lets add a:

```
g_debug ("Failed to auto-detect branch name from git");
```

src/builder-source-git.c
```diff
@@ -478,4 +486,5 @@ builder_source_git_class_init (BuilderSourceGitClass *klass)
 static void
 builder_source_git_init (BuilderSourceGit *self)
 {
+  self->default_branch_name = NULL;
```
TingPing, line 489:
```suggestion
```

The entire object should be 0 initialized so this isn't needed.

PR #459, opened 2022-01-31
source-archive, source-file: Deprecate MD5 and SHA1 hashes

debarshiray:
... because of their weaknesses, and show a warning suggesting SHA512
instead. SHA256 is another good option, but presumably SHA512 is more
future-proof than SHA256.


debarshiray:
I chose the deprecation notice to be something that fits within 80 characters and is self-documenting, but I am open to suggestions.

smcv:
> SHA256 is another good option, but presumably SHA512 is more future-proof than SHA256

The obvious disadvantage is that SHA512 hashes are twice as long. It'll be more resistant to brute-force attacks, but that might be a matter of "it only takes *half* the age of the universe to carry out a successful brute-force"...

If there's a practical faster-than-brute-force attack on the SHA-2 family, it's not clear that SHA512 will be any less vulnerable to it than the other members of the family.

Also, SHA256 is used in libostree (therefore Flatpak) for its own integrity checks, so if SHA256 has practical problems, we have plenty of work to do to replace it, regardless.

debarshiray:
My vague understanding is that, in practice, MD5 and SHA1 hashes are only used for generated dependencies from other distribution systems like Maven, NPM and Yarn.  So, I was wondering whether it will be prudent to use a thicker stick to prod those ecosystems to move to a better hash.

smcv:
Oh, I'm not saying that deprecating MD5 and SHA1 is bad - just that SHA256 might be a more appropriate recommendation than 512.

debarshiray:
> Oh, I'm not saying that deprecating MD5 and SHA1 is bad - just that
> SHA256 might be a more appropriate recommendation than 512.

Yes, I understood.  I meant that since it might take a long time to get ecosystems like Maven, NPM and Yarn to migrate, I was wondering if it might be a good idea to push them all the way to SHA512, just to be sure.  :)

Anyway, I changed the code to suggest SHA256.

debarshiray:
Ping.

debarshiray:
Ping.

smcv (approved):

PR #464, opened 2022-04-01
[flatpak-builder-1.0.x] Backport CVE-2022-21682 fixes

debarshiray:
I am going to include these in RHEL 8.7.  Given that these are security fixes, I thought that I might as well put them on the upstream `flatpak-builder-1.0.x` branch, and let everybody benefit.

----

* Disable filesystem access with --nofilesystem=host:reset
  (cherry picked from commit 2d1b6799e782d5e5577072aa7bbfed00ddf0b087)

* Allow --nofilesystem=host:reset in flatpak-builder --run
  (cherry picked from commit 27aa2f5e4508d03178bf905ee7099d6d69a79aa4)


smcv (approved):
This seems correct, is mostly backwards-compatible with old Flatpak versions (the `host:reset` syntax was chosen to show warnings but otherwise work), and matches what I applied in Debian 11.

I don't think we plan to do any more 1.0.x releases (unless @alexlarsson or @mwleeds want to do one?) so we should probably also add a `SECURITY.md` clarifying support status, similar to what we did in Flatpak itself. For Flatpak, we're security-supporting more than one branch, but for flatpak-builder I think we are not.

debarshiray:
Thanks for the review, @smcv !

PR #470, opened 2022-05-10
Add option to override SOURCE_DATE_EPOCH.

zaclegarssure:
`flatpak-builder` overrides `SOURCE_DATE_EPOCH` with the last modification time of the manifest file (of the program being built).
With this option, it would provide a more convenient interface to perform the same thing, useful when trying to reproduce a build.
Concretely I just added an optional parameter `--override-source-date-epoch` which, if set, will override the `SOURCE_DATE_EPOCH` with the given value. 
It's my first time looking at the code of this project, therefore, even though I made minimal changes, I might have made some mistakes.
Update:
I also added the timestamp to the resolved manifest, to actually have reproducible builds.

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -36,6 +36,7 @@
 
 #include <libxml/parser.h>
 
+#include "glibconfig.h"
 #include "libglnx/libglnx.h"
 
 #define LOCALES_SEPARATE_DIR "share/runtime/locale"
@@ -107,6 +108,7 @@ struct BuilderManifest
   GList          *expanded_modules;
   GList          *add_extensions;
   GList          *add_build_extensions;
+  gint64          source_date_epoch;
 };
 
 typedef struct
@@ -169,6 +171,7 @@ enum {
   PROP_ADD_BUILD_EXTENSIONS,
   PROP_EXTENSION_TAG,
   PROP_TOKEN_TYPE,
+  PROP_SOURCE_DATE_EPOCH,
   LAST_PROP
 };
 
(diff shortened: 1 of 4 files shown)


src/builder-manifest.c
```diff
@@ -1090,6 +1100,16 @@ builder_manifest_class_init (BuilderManifestClass *klass)
...
+
+  g_object_class_install_property (object_class,
+                                   PROP_SOURCE_DATE_EPOCH,
+                                   g_param_spec_int64 ("source-date-epoch",
+                                                     "",
+                                                     "",
+                                                     0,
+                                                     G_MAXINT64,
+                                                     0,
+                                                     G_PARAM_READWRITE));
```
TingPing, line 1112:
```suggestion
                                                      "",
                                                      "",
                                                      0,
                                                      G_MAXINT64,
                                                      0,
                                                      G_PARAM_READWRITE));
```

TingPing: Hard to tell on the github UI, but alignment was just off by one or two.

src/builder-manifest.c
```diff
@@ -1629,6 +1649,8 @@ builder_manifest_start (BuilderManifest *self,
   g_autofree char *sdk_path = NULL;
   const char *stop_at;
 
+  self->source_date_epoch = builder_context_get_source_date_epoch(context);
```
TingPing, line 1652:
```suggestion
  self->source_date_epoch = builder_context_get_source_date_epoch (context);
```

TingPing:
The change itself sounds and looks good to me.

src/builder-manifest.c
TingPing: It was off by one more space sorry :/

zaclegarssure: No problems, I should have been more careful in the first place.

PR #481, opened 2022-07-20
Add sdk-version

JakobDev:
If you want to specify another SDK version than the runtime (what is needed e.g. for extensions) you currently have to use `sdk: name//version`. We already have things like `runtime-version` and `base-version`, so why not add `sdk-version`?

The code is (of course) tested and works. I hope I didn't miss anything.

### DIFF
--- a/data/flatpak-manifest.schema.json
+++ b/data/flatpak-manifest.schema.json
@@ -1155,6 +1155,10 @@
       "description": "The name of the development runtime that the application builds with.",
       "type": "string"
     },
+    "sdk-version": {
+      "description": "The version of the development runtime that the application builds with, defaults to the runtime version.",
+      "type": "string"
+    },
     "var": {
       "description": "Initialize the (otherwise empty) writable /var in the build with a copy of this runtime.",
       "type": "string"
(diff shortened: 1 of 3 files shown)


Erick555:
Which extensions needs different Sdk version?

JakobDev:
Extensions use the App the Extension is for as Runtime and a normal SDK. Because the SDK has not the same Version as the App, you need to specify the version. See [this OBS plugin](https://github.com/flathub/com.obsproject.Studio.Plugin.MoveTransition/blob/master/com.obsproject.Studio.Plugin.MoveTransition.yaml) for example.

Erick555:
This looks quite hacky. Parent app runtime doesn't necessarily have to be abi compatible with child Sdk. I'm not sure if flatpak should officially acknowledge this usecase.

Erick555:
Ok. The problem I see now is it's backwards incompatible change - all older f-b would fail to properly build new manifest. Maybe it's not worth it if there is easy workaround.

Erick555:
The point is: why bother? This fixes pure cosmetic issue. I doubt adjusting manifest for this is worth anyone time.

JakobDev:
The point here is, that on every point in the manifest you can set a Flatpak ID you also have a separate Key for a version. The SDK is the only point where you need the `id//version` format. So missing this here could be considered as a Bug.

Erick555:
I could agree yet not every bug is worth fixing.

Issue #477, opened 2022-06-20
[Feature request]: Ability to build without libsoup2


nanonyme:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

Please either allow building with libsoup2 replaced with libsoup3 or just libcurl. libsoup2 is going away.

nanonyme:
In fact looks a lot like this project was already ported to curl but proxy usage was left to libsoup instead of using glib tooling.

nanonyme:
Flatpak itself gained ability not to use libsoup2 through https://github.com/flatpak/flatpak/pull/4943

nanonyme:
If the only reason is proxy, something like https://github.com/hughsie/appstream-glib/pull/443 would probably be enough.

nanonyme:
Looks like also SoupURI is used.

nanonyme:
builder_source_file_extract codepath is clearly directly using libsoup2.

PR #491, opened 2022-09-02
Get rid of remainder of libsoup usage in flatpak-builder

nanonyme:
Fixes https://github.com/flatpak/flatpak-builder/issues/477

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -11,12 +11,10 @@ on:
 jobs:
   check:
     name: Build with gcc and test
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
     - name: Install Dependencies
       run: |
-        sudo add-apt-repository ppa:alexlarsson/flatpak
-        sudo add-apt-repository 'deb https://download.mono-project.com/repo/ubuntu stable-focal main' # Needed for updates to work
         sudo apt-get update
         sudo apt-get install -y libglib2.0 attr automake gettext autopoint bison  dbus gtk-doc-tools \
         libfuse-dev ostree libostree-dev libarchive-dev libcap-dev libattr1-dev libdw-dev libelf-dev \
@@ -42,12 +40,10 @@ jobs:
 
   clang:
     name: Build with clang
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
     - name: Install Dependencies
       run: |
-        sudo add-apt-repository ppa:alexlarsson/flatpak
-        sudo add-apt-repository 'deb https://download.mono-project.com/repo/ubuntu stable-focal main' # Needed for updates to work
         sudo apt-get update
         sudo apt-get install -y libglib2.0 attr automake gettext autopoint bison  dbus gtk-doc-tools \
         libfuse-dev ostree libostree-dev libarchive-dev libcap-dev libattr1-dev libdw-dev libelf-dev \
--- a/configure.ac
+++ b/configure.ac
@@ -12,7 +12,7 @@ AC_INIT([flatpak-builder],
         [flatpak-builder],
         [http://flatpak.org/])
 
-GLIB_REQS=2.44
+GLIB_REQS=2.66
 OSTREE_REQS=2017.14
 FLATPAK_REQS=0.99.1
 SYSTEM_DEBUGEDIT_REQS=5.0
@@ -83,7 +83,7 @@ PKG_PROG_PKG_CONFIG([0.24])
 # For libglnx
 AC_CHECK_HEADER([sys/xattr.h], [], [AC_MSG_ERROR([You must have sys/xattr.h from glibc])])
 
-PKG_CHECK_MODULES(BASE, [glib-2.0 >= $GLIB_REQS gio-2.0 gio-unix-2.0 libsoup-2.4 ostree-1 >= $OSTREE_REQS json-glib-1.0 libxml-2.0 >= 2.4 libcurl])
+PKG_CHECK_MODULES(BASE, [glib-2.0 >= $GLIB_REQS gio-2.0 gio-unix-2.0 ostree-1 >= $OSTREE_REQS json-glib-1.0 libxml-2.0 >= 2.4 libcurl])
 
 dnl ************************
 dnl *** check for libelf ***
--- a/src/builder-context.c
+++ b/src/builder-context.c
@@ -37,6 +37,7 @@
 #include "builder-cache.h"
 #include "builder-utils.h"
 
+
 struct BuilderContext
 {
   GObject         parent;
@@ -45,7 +46,6 @@ struct BuilderContext
   GFile          *run_dir; /* directory flatpak-builder was started from */
   GFile          *base_dir; /* directory with json manifest, origin for source files */
   char           *state_subdir;
-  SoupSession    *soup_session;
   CURL           *curl_session;
   char           *arch;
   char           *default_branch;
@@ -116,7 +116,6 @@ builder_context_finalize (GObject *object)
   g_clear_object (&self->app_dir);
   g_clear_object (&self->run_dir);
   g_clear_object (&self->base_dir);
-  g_clear_object (&self->soup_session);
   g_clear_object (&self->options);
   g_clear_object (&self->sdk_config);
   g_free (self->arch);
@@ -383,24 +382,32 @@ builder_context_download_uri (BuilderContext *self,
                               GError        **error)
 {
   int i;
-  g_autoptr(SoupURI) original_uri = soup_uri_new (url);
   g_autoptr(GError) first_error = NULL;
+  g_autoptr(GUri) original_uri = g_uri_parse (url, CONTEXT_HTTP_URI_FLAGS, &first_error);
 
   if (original_uri == NULL)
-    return flatpak_fail (error, _("Could not parse URI “%s”"), url);
+    {
+      g_propagate_error (error, g_steal_pointer (&first_error));
+      return FALSE;
+    }
 
   g_print ("Downloading %s\n", url);
 
   if (self->sources_urls != NULL)
     {
-      g_autofree char *base_name = g_path_get_basename (soup_uri_get_path (original_uri));
+      g_autofree char *base_name = g_path_get_basename (g_uri_get_path (original_uri));
       g_autofree char *rel = g_build_filename ("downloads", checksums[0], base_name, NULL);
 
       for (i = 0; i < self->sources_urls->len; i++)
         {
-          SoupURI *base_uri = g_ptr_array_index (self->sources_urls, i);
-          g_autoptr(SoupURI) mirror_uri = soup_uri_new_with_base (base_uri, rel);
-          g_autofree char *mirror_uri_str = soup_uri_to_string (mirror_uri, FALSE);
+          GUri *base_uri = g_ptr_array_index (self->sources_urls, i);
+          g_autoptr(GUri) mirror_uri = g_uri_parse_relative (base_uri, rel, CONTEXT_HTTP_URI_FLAGS, &first_error);
+          if (mirror_uri == NULL)
+            {
+              g_propagate_error (error, g_steal_pointer (&first_error));
+              return FALSE;
+            }
+          g_autofree char *mirror_uri_str = g_uri_to_string (mirror_uri);
           g_print ("Trying mirror %s\n", mirror_uri_str);
           g_autoptr(GError) my_error = NULL;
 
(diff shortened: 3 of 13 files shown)


nanonyme:
@alexlarsson @TingPing this is probably awful stylically and haven't yet run this through valgrind but before continuing need decision what to do with glib compat. I am still working on validating everything works. Does flatpak-builder need to keep supporting glib 2.44? If it can be bumped to 2.66 (involves bumping builders to 22.04), we don't need compat layer in flatpak-builder which keeps code simpler.

src/builder-context.h
```diff
@@ -22,14 +22,16 @@
...
-#include <libsoup/soup.h>
 #include <curl/curl.h>
 #include "builder-options.h"
 #include "builder-utils.h"
 #include "builder-sdk-config.h"
 
 G_BEGIN_DECLS
 
+/* Same as SOUP_HTTP_URI_FLAGS, means all possible flags for http uris */
+#define CONTEXT_HTTP_URI_FLAGS (G_URI_FLAGS_HAS_PASSWORD | G_URI_FLAGS_ENCODED_PATH | G_URI_FLAGS_ENCODED_QUERY | G_URI_FLAGS_ENCODED_FRAGMENT | G_URI_FLAGS_SCHEME_NORMALIZE)
```
smcv, line 33: `GUri` with `G_URI_FLAGS_SCHEME_NORMALIZE` requires bumping `GLIB_REQS` in `configure.ac` to 2.68 (see flatpak/flatpak#5062 and flatpak/flatpak#5064), which would rule out compiling upstream source on Debian 11.

Alternatively, if it's OK to use `G_URI_FLAGS_SCHEME_NORMALIZE` if and only if it exists (as in flatpak/flatpak#5064), then that use of `GUri` will depend on GLib 2.66, which is available in Debian 11.

smcv (changes requested):

src/builder-context.h
nanonyme: Yeah, I know the failure reason. Trying to raise some discussion on it. It is clear flatpak needs to support ancient distros but with flatpak-builder situation has been more fuzzy.

nanonyme: But yeah, maybe not raising bar above 2.66 would be a good call anyway. This build runs with Ubuntu 20.04 which I suppose has even older glib.

nanonyme: Right. I now did this change, there's obvious error for logs `Requested 'glib-2.0 >= 2.66' but version of GLib is 2.64.6`

nanonyme: Does flatpak require glib 2.66 to build?

smcv: Flatpak itself has a backport of GUri for GLib older than 2.66. The problem in flatpak/flatpak#5062 is that if GLib is *exactly* 2.66, as in Debian 11, the backport isn't used but the real OS GLib doesn't support `G_URI_FLAGS_SCHEME_NORMALIZE`.

smcv: > It is clear flatpak needs to support ancient distros but with flatpak-builder situation has been more fuzzy

Flatpak allegedly supports ancient distros, but it doesn't *actually* support (sufficiently) ancient distros: flatpak/flatpak#5066.

For flatpak-builder and other ancillary stuff, I personally think a reasonable place to draw the line is the latest Debian stable release or the latest Ubuntu LTS, whichever is older. They're both released on a cycle of about 2 years, with an offset of 1 year, so the one that is the limiting factor swaps periodically: it's currently Debian 11, but when Debian 12 is released, Ubuntu 22.04 will become the limiting factor.

For Flatpak itself, clearly there's interest in supporting older distros in order to be able to use Flatpak to run recent software on old distros, and in particular people do use it on older-than-latest-LTS branches of Ubuntu (although I'm the only one who is actually working on those backports, and I don't use Ubuntu myself). I think we should seriously consider dropping any attempt to backport to Ubuntu 16.04 (which is EOL for Ubuntu's standard support, and requires a support subscription for extended security maintenance), and perhaps also 18.04 (which is not EOL, but Flatpak doesn't compile successfully there and nobody noticed until I tried to update the backport).

nanonyme: Seems there is no flatpak backports for jammy.

nanonyme: I did not touch that part of build that did `sudo add-apt-repository ppa:alexlarsson/flatpak` yet.

nanonyme: Appears build is also doing `sudo add-apt-repository 'deb https://download.mono-project.com/repo/ubuntu stable-jammy main' # Needed for updates to work`. I updated name, seems there is no package for jammy. I don't know what we need from there.

smcv: Adding the Mono repository was a workaround for a bug in how Github set up their containers: see <https://github.com/flatpak/flatpak/pull/4299>. If we don't need to use that workaround any more in 22.04, then it's better to stop using it (remove that line). We only want that workaround if it's still necessary.

smcv: It's probably actually better if we do each CI build with a minimum of backports, and only add extra non-Ubuntu repositories as we find that we need them. Ubuntu 22.04 has Flatpak 1.12.x, which seems like it should be sufficient for flatpak-builder for a while.

nanonyme: Looks like Python 2 test is skipped completely currently since build machine does not have Python 2 installed.

nanonyme: @smcv I thought I did the requested changes but GitHub is still complaining. Did I miss something?

.github/workflows/check.yml
```diff
@@ -11,12 +11,10 @@ on:
 jobs:
   check:
     name: Build with gcc and test
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
TingPing, line 14: This can be a separate commit.

configure.ac
```diff
@@ -12,7 +12,7 @@ AC_INIT([flatpak-builder],
         [flatpak-builder],
         [http://flatpak.org/])
 
-GLIB_REQS=2.44
+GLIB_REQS=2.66
```
TingPing, line 15: This is a massive bump. I don't think its acceptable.

smcv: What older OSs are you interested in running flatpak-builder on? When there's a GLib feature that we can benefit from as much as the existence of GUri, I think the required version should be driven by which OSs we want to support, rather than by backwards-compat for the sake of backwards-compat.

gasinvein: It's flatpak-builder, not flatpak. Do we really need to support _building_ flatpaks on something older than Debian Stable? There is `org.flatpak.Builder` flatpak after all.

.github/workflows/check.yml
nanonyme: Okay, something to keep in mind these changes are interdependent. Glib requires builder bump. But I can split.

configure.ac
TingPing: I didn't track down who has it. Debian Stable having it is a reasonable metric for this being OK.

.github/workflows/check.yml
nanonyme: I have now split it so that CI is updated as first commit and this work is on top of that.

nanonyme:
So are there still further changes requested?

src/builder-context.c
```diff
@@ -383,24 +382,31 @@ builder_context_download_uri (BuilderContext *self,
...
-  g_autoptr(SoupURI) original_uri = soup_uri_new (url);
   g_autoptr(GError) first_error = NULL;
+  g_autoptr(GUri) original_uri = g_uri_parse (url, CONTEXT_HTTP_URI_FLAGS, &first_error);
 
-  if (original_uri == NULL)
-    return flatpak_fail (error, _("Could not parse URI “%s”"), url);
+  if (original_uri == NULL) {
+    g_propagate_error (error, g_steal_pointer (&first_error));
+    return FALSE;
+  }
```
smcv, line 391:
```
  if (original_uri == NULL)
    {
      g_propagate_error (error, g_steal_pointer (&first_error));
      return FALSE;
    }
```

src/builder-context.h
```diff
@@ -22,14 +22,22 @@
...
 #include <curl/curl.h>
 #include "builder-options.h"
 #include "builder-utils.h"
 #include "builder-sdk-config.h"
 
 G_BEGIN_DECLS
 
+/* Same as SOUP_HTTP_URI_FLAGS, means all possible flags for http uris */
+
+#if GLIB_CHECK_VERSION (2, 68, 0) || !GLIB_CHECK_VERSION (2, 66, 0)
```
smcv, line 34:
```
#if GLIB_CHECK_VERSION (2, 68, 0)
```

Unlike Flatpak, flatpak-builder no longer supports versions older than 2.66 after this commit, and doesn't have a backport of GUri.

src/builder-main.c
```diff
@@ -553,7 +545,7 @@ main (int    argc,
   if (opt_sources_urls)
     {
       g_autoptr(GPtrArray) sources_urls = NULL;
-      sources_urls = g_ptr_array_new_with_free_func ((GDestroyNotify)soup_uri_free);
+      sources_urls = g_ptr_array_new_with_free_func ((GDestroyNotify)g_free);
```
smcv, line 548:
```
      sources_urls = g_ptr_array_new_with_free_func ((GDestroyNotify) g_uri_unref);
```

src/builder-main.c
```diff
@@ -562,7 +554,7 @@ main (int    argc,
               opt_sources_urls[i] = g_strdup_printf ("%s/", tmp);
             }
 
-          SoupURI *uri = soup_uri_new (opt_sources_urls[i]);
+          GUri *uri = g_uri_parse (opt_sources_urls[i], CONTEXT_HTTP_URI_FLAGS, &error);
           if (uri == NULL)
             {
               g_printerr ("Invalid URL '%s'", opt_sources_urls[i]);
```
smcv, line 560: You probably want to show `error->message` here?

src/builder-source-file.c
```diff
@@ -329,23 +328,26 @@ download_data_uri (const char     *url,
...
+  g_autoptr(GError) first_error = NULL;  
   g_autoptr(GInputStream) input = NULL;
   g_autoptr(GOutputStream) out = NULL;
+  g_autoptr(GUri) parsed = NULL;
 
-  session = builder_context_get_soup_session (context);
+  parsed = g_uri_parse(url, CONTEXT_HTTP_URI_FLAGS, error);
+  if (!parsed) {
+    return FALSE;
+  }
```
smcv, line 340:
```
  if (!parsed)
    return FALSE;
```

or

```
  if (!parsed)
    {
      return FALSE;
    }
```

src/builder-source-file.c
```diff
@@ -329,23 +328,26 @@ download_data_uri (const char     *url,
...
+  session = builder_context_get_curl_session (context);
+  out = g_memory_output_stream_new_resizable ();
 
-  input = soup_request_send (req, NULL, error);
-  if (input == NULL)
+  if (!builder_download_uri_buffer(parsed, session, out, NULL, 0, error))
+    {
+    g_propagate_error (error, g_steal_pointer (&first_error));
     return NULL;
+    }
```
smcv, line 349:
```
    {
      g_propagate_error (error, g_steal_pointer (&first_error));
      return NULL;
    }
```

src/builder-context.h
```diff
@@ -22,14 +22,22 @@
...
+#define CONTEXT_HTTP_URI_FLAGS (G_URI_FLAGS_HAS_PASSWORD | G_URI_FLAGS_ENCODED_PATH | G_URI_FLAGS_ENCODED_QUERY | G_URI_FLAGS_ENCODED_FRAGMENT | G_URI_FLAGS_SCHEME_NORMALIZE)
+#else
+/* GLib 2.66 didn't support scheme-based normalization */
```
TingPing, line 37: We don't actually care unless we are doing comparisons between URIs. (We may or may not, I haven't looked).

src/builder-source-file.c
```diff
@@ -329,23 +328,26 @@ download_data_uri (const char     *url,
...
 
-  req = soup_session_request (session, url, error);
-  if (req == NULL)
-    return NULL;
+  session = builder_context_get_curl_session (context);
+  out = g_memory_output_stream_new_resizable ();
 
-  input = soup_request_send (req, NULL, error);
-  if (input == NULL)
+  if (!builder_download_uri_buffer(parsed, session, out, NULL, 0, error))
```
TingPing, line 345:
```suggestion
  if (!builder_download_uri_buffer (parsed, session, out, NULL, 0, error))
```

src/builder-context.h
TingPing, line 34: I'm not saying we care but its worth noting CURL has a complete API to parse URIs and GUri does nothing unique for us.

But of course the CURL API is the CURL API so its a lot uglier than GUri. https://curl.se/libcurl/c/libcurl-url.html

nanonyme: Considering large amounts of this code is almost copy-paste from flatpak porting away from libsoup2, I would prefer not going into a larger refactoring into CURL API which I don't know well.

src/builder-context.h
nanonyme: Sure, this was mostly copy-paste from flatpak which did similar migration earlier. Afaik the thought behind is that with at least newer versions of glib you get behaviour that is equivalent to the old behaviour.

TingPing (approved):
This seems good to me, is it ready to merge @nanonyme ?

nanonyme:
Well, I was trying it out to build various apps, didn't notice anything weird. There's a risk I didn't have complex enough apps to test through all possible scenarios.

PR #498, opened 2022-11-16
Checkout LFS objects

barthalion:


Doomsdayrs:
Has this commit been merged or patched into the flathub buildbot? Since I am seeing its output here: https://buildbot.flathub.org/#/builders/6/builds/128989

bbhtt:
org.flatpak.Builder in Flathub uses it as a patch, but it doesn't work anymore.

bbhtt:
Seems the patch still works, the issue is with your git repo https://github.com/flathub/com.revolutionarygamesstudio.ThriveLauncher/pull/50#issuecomment-2177354958

PR #504, opened 2022-12-17
Replace deprecated functions

krieghof:
Fixes compiler warnings about deprecated functions.


src/builder-utils.c
```diff
@@ -883,7 +883,7 @@ builder_host_spawnv (GFile                *dir,
   g_source_remove (sigint_id);
   g_dbus_connection_signal_unsubscribe (connection, subscription);
 
-  if (!g_spawn_check_exit_status (data.exit_status, error))
+  if (!g_spawn_check_wait_status (data.exit_status, error))
```
TingPing, line 886: This is a much newer API, we probably don't want to require this version of GLIB.

Instead what we should do is define `-DGLIB_VERSION_MIN_REQUIRED=GLIB_VERSION_2_44 -DGLIB_VERSION_MAX_ALLOWED=GLIB_VERSION_2_44` in our build system. This will silence the deprecation warnings.

src/builder-main.c
```diff
@@ -430,7 +430,7 @@ main (int    argc,
   g_type_ensure (G_TYPE_SOCKET_PROTOCOL);
   g_type_ensure (G_TYPE_SOCKET_ADDRESS);
 
-  orig_argv = g_memdup (argv, sizeof (char *) * argc);
+  orig_argv = g_memdup2 (argv, sizeof (char *) * argc);
```
TingPing, line 433: We can do this without bumping any requirements, however you need to update the `libglnx` submodule and use its header.

src/builder-context.c
```diff
@@ -900,9 +900,9 @@ builder_context_enable_rofiles (BuilderContext *self,
           new_action.sa_flags = 0;
           sigaction (SIGHUP, &new_action, NULL);
 
-          sigset (SIGINT, SIG_IGN);
-          sigset (SIGPIPE, SIG_IGN);
-          sigset (SIGSTOP, SIG_IGN);
+          signal (SIGINT, SIG_IGN);
+          signal (SIGPIPE, SIG_IGN);
+          signal (SIGSTOP, SIG_IGN);
```
TingPing, line 905: I'm not really familiar with signal APIs but this seems like the wrong direction?

My understanding is `sigset` was the newer, better defined, API to use over `signal`. Maybe a newer API replaces this, or do you have more context?

TingPing (changes requested):

src/builder-context.c
TingPing: The docs say:

> This API is obsolete: new applications should use the POSIX signal API ([sigaction](https://linux.die.net/man/2/sigaction)(2), [sigprocmask](https://linux.die.net/man/2/sigprocmask)(2), etc.)

So I think one of those APIs should be used.

krieghof: And here we have the deprecated msg from the header :laughing: 

``` 
/* Set the disposition of SIG.  */
extern __sighandler_t sigset (int __sig, __sighandler_t __disp) __THROW
  __attribute_deprecated_msg__
    ("Use the signal and sigprocmask functions instead");
#endif
```

But I like the idea with `sigaction` also better, especially because   we already use it in that scope.

src/builder-utils.c
krieghof:
```
AC_DEFINE([GLIB_VERSION_MIN_REQUIRED], [GLIB_VERSION_2_66], [Glib min version])
AC_DEFINE([GLIB_VERSION_MAX_ALLOWED], [GLIB_VERSION_2_66], [Glib max version])
```

Defining those two macros result in an 'Not available before 2.68' warning for `g_memdup2`. Maybe something we can fix in libglnx?

But it also throws the same warning for `G_URI_FLAGS_SCHEME_NORMALIZE`, that got just introduced with the libsoup changes, but the min version got only bumped to version 2.66. I couldn't find any signs of a backport therefore I think we should increase the version to 2.68. 

If that's the case should I remove the commit with the libglnx bump?

TingPing: > But it also throws the same warning for G_URI_FLAGS_SCHEME_NORMALIZE, that got just introduced with the libsoup changes, but the min version got only bumped to version 2.66.

It is in a `#if GLIB_CHECK_VERSION (2, 68, 0)` check block so its fine.

Let's just do `GLIB_VERSION_MAX_ALLOWED=GLIB_VERSION_2_68` and leave the MIN.

krieghof: Now I see, thanks.

Leaving the min version completely out didn't work, so I set it to 2.66.

But maybe it is better to just drop the changes, the commit complicates the build config only to silence a compiler warning. If setting those macros doesn't have any other benefits I'm fine with dropping this commit. What do you think?

TingPing: The macros are great! It will prevent future slip ups when using too new of APIs.

PR #506, opened 2022-12-26
builder-manifest: Report the index of the module with the missing name

krieghof:
This should help the user identify the broken module more easily.

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -225,8 +225,9 @@ expand_modules (BuilderContext *context, GList *modules,
                 GList **expanded, GHashTable *names, GError **error)
 {
   GList *l;
+  gsize i = 1;
 
-  for (l = modules; l; l = l->next)
+  for (l = modules; l; l = l->next, i++)
     {
       BuilderModule *m = l->data;
       GList *submodules = NULL;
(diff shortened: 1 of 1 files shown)


src/builder-manifest.c
```diff
@@ -246,11 +247,8 @@ expand_modules (BuilderContext *context, GList *modules,
       name = builder_module_get_name (m);
       if (name == NULL)
         {
-          /* FIXME: We'd like to report *something* for the user
-                    to locate the errornous module definition.
-           */
           g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
-                       "Module has no 'name' attribute set");
+                       "The %lu. Module has no 'name' attribute set", i);
```
TingPing, line 251: This phrasing is odd "The 3. Module has no...".

Maybe "Module (3) has no..."

However I have another concern which is that this isn't helpful for any manifest with submodules, as it will point you in the wrong place.

hfiguiere: If I'm not mistaken the `.` after the number is the German equivalent of `th` in English.

krieghof: @hfiguiere Yeah, that's right I'm from Austria and we use the `.` over here. 
Unfortunately, I'm pretty  busy for the next month, and it will take some time until I can try to fix the problem with the submodules and the bad grammar.

PR #508, opened 2023-01-02
Run shell before the start of the build

### LINKED ISSUE #479: [Bug]: --build-shell parameter stops after configuration not download/unpackaging
### Checklist

- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a bug that matches the one I want to file, without success.

### flatpak-builder version

1.2.2

### Flatpak version

1.12.7

### How to reproduce

Create a module that fails in the configuration step

### Expected Behavior

Just as described in the documentation for --build-shell  

"Extract and prepare the sources for the named module,
and then start a shell in a sandbox identical to the one flatpak-builder
would use for building the module. This is useful to debug a module."

### Actual Behavior

It doesnt stop after downloading and extracting the sources, but 
AFTER  the configuration step. And if the configuration step fails there is no
way to get into the build shell and see what the problem is 

Either this or the documentation is wrong and this parameter loses some of it's usefullness.

### Additional Information

_No response_

koko-ng:
Hi :)
For some build systems there were calls to `build` before `shell`, which made it impossible to open a shell in the build environment if the build failed. I don't think that moving `shell` before has an implication on the environment as the code in between only seems to do checks on the project.

Fixes #479

### DIFF
--- a/src/builder-module.c
+++ b/src/builder-module.c
@@ -1648,6 +1648,13 @@ builder_module_build_helper (BuilderModule  *self,
         }
     }
 
+  if (run_shell)
+    {
+      if (!shell (app_dir, self->name, context, source_dir, build_dir_relative, build_args, env, error))
+        return FALSE;
+      return TRUE;
+    }
+
   if (configure_file)
     has_configure = g_file_query_exists (configure_file, NULL);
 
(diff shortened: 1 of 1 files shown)


TingPing:
It does change the behavior since previously it would configure the project, then open a shell, now it just opens a shell to an empty directory.

I'm not sure this distinction matters or not but another possibility is if `run_shell` is `true` to not consider configuration failures as fatal?

nbenitez:
> I'm not sure this distinction matters or not but another possibility is if run_shell is TRUE to not consider configuration failures as fatal?

Yes please, otherwise how am I suppose to debug a module that fails in configure step? if I'm not able to open a shell in that case..

PR #510, opened 2023-01-09
git: Just skip submodules that git ls-tree don't know about

hfiguiere:
I found this bug with this repository:

https://github.com/GuitarML/SmartGuitarAmp

For the resulting flatpak manifest I had to skip submodules and add the sources.

See https://github.com/hfiguiere/flathub/blob/guitarml/org.freedesktop.LinuxAudio.Plugins.GuitarML.json#L231

So basically if git ls-tree return nothing when the submodule is list in gitmodule it will just be skipped. I check with my manifest and could successfully reenable the submodules.

### DIFF
--- a/src/builder-git.c
+++ b/src/builder-git.c
@@ -395,11 +395,9 @@ git_mirror_submodules (const char     *repo_location,
             return FALSE;
 
           lines = g_strsplit (g_strstrip (ls_tree), "\n", 0);
+          /* There can be path in the .gitmodules file that are not submodules. */
           if (g_strv_length (lines) != 1)
-            {
-              g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "Not a gitlink tree: %s", path);
-              return FALSE;
-            }
+            continue;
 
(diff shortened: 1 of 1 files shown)


src/builder-git.c
```diff
@@ -396,10 +396,8 @@ git_mirror_submodules (const char     *repo_location,
 
           lines = g_strsplit (g_strstrip (ls_tree), "\n", 0);
           if (g_strv_length (lines) != 1)
-            {
-              g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "Not a gitlink tree: %s", path);
-              return FALSE;
-            }
+            // There can be path in the gitmodules files that are not submodules.
+            continue;
```
TingPing, line 400: Just minor nits: Try to follow the style around this.

```c
/* There can be a path in the gitmodules files thare are not submodules. */
if (g_strv_length (lines) != 1)
  continue;
```

hfiguiere: done (and fixed grammar).

Issue #511, opened 2023-01-25
[Bug]: Some test data missing from 1.3.1 tarball


smcv:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a bug that matches the one I want to file, without success.

### flatpak-builder version

1.3.1

### Flatpak version

1.14.1

### How to reproduce

* Build for OS packages with as-installed tests enabled (`--prefix=/usr --enable-installed-tests` etc., as in the Debian packaging)
* Install
* Run tests with `ginsttest-runner`

### Expected Behavior

Tests pass

### Actual Behavior

```
Running test: flatpak-builder/test-builder-deprecated.sh.test
1..8
...
cp: cannot stat '/usr/libexec/installed-tests/flatpak-builder/org.test.Deprecated.SHA1.archive.yaml': No such file or directory
OK closing connection
FAIL: flatpak-builder/test-builder-deprecated.sh.test (Child process exited with code 1)
```

### Additional Information

Merge request to follow.

PR #512, opened 2023-01-25
Fix installation of "as-installed" tests

smcv:
* tests: Don't install buildutil/tap-driver.sh
    
    This is not necessary for the "as-installed" tests, and happens to cause
    lint warnings from Debian's Lintian tool.

* tests: Install all deprecation tests when installed-tests are enabled

    Resolves: #511


smcv:
> Tests are still running, but I think this is right.

Yes, seems to work.

PR #517, opened 2023-02-03
Port to libappstream

Lunarequest:
This the completed port to libappstream with a counterpart on the flatpak side(https://github.com/flatpak/flatpak/pull/5277). It is based of off @pwithnall and @barthalion previous work. 
Reasoning has been taken from @pwithnall's original pr
> appstream-glib is mostly unmaintained, and appstream is more
actively developed (and up to date with the AppStream specification).

> This ports flatpak-builder to use the command line tools from appstream
(appstreamcli and appstreamcli-compose) rather than the ones from
appstream-glib (appstream-util, appstream-builder and
appstream-compose).

> This should result in no changes for most things, but should fix
propagation of \<requires>/\<recommends> elements, screenshot
>\<caption>s, and other elements which have been added to the AppStream
specification recently.

> For example, see https://github.com/flathub/flathub/issues/2439

Signed-off-by: Luna D Dragon [luna.dragon@suse.com](mailto:luna.dragon@suse.com)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -18,6 +18,7 @@ AC_DEFINE([GLIB_VERSION_MAX_ALLOWED], [GLIB_VERSION_2_68], [Glib max version])
 GLIB_REQS=2.66
 OSTREE_REQS=2017.14
 FLATPAK_REQS=0.99.1
+APPSTREAMCLI_REQS=0.15.0
 SYSTEM_DEBUGEDIT_REQS=5.0
 LIBDW_REQS=0.172
 
@@ -40,6 +41,21 @@ else
                       [AC_MSG_ERROR([You need at least version $FLATPAK_REQS of flatpak, your version is $FLATPAK_VERSION])])
 fi
 
+AC_CHECK_PROG([APPSTREAMCLI], [appstreamcli], [appstreamcli], [false])
+if test "x$APPSTREAMCLI" = xfalse; then
+   AC_MSG_ERROR([You need appstreamcli installed])
+   APPSTREAMCLI_VERSION=`$APPSTREAMCLI --version | sed 's,.*\ \([0-9]*\.[0-9]*\.[0-9]*\)$,\1,'`
+   AX_COMPARE_VERSION([$APPSTREAMCLI_REQS],[gt],[$APPSTREAMCLI_VERSION],
+                      [AC_MSG_ERROR([You need at least version $APPSTREAMCLI_REQS of appstreamcli, your version is $APPSTREAMCLI_VERSION])])
+fi
+
+AC_MSG_CHECKING([whether appstreamcli has compose support])
+AS_IF([appstreamcli compose --help >/dev/null 2>&1],
+      [AC_MSG_RESULT(yes)],
+      [
+       AC_MSG_RESULT(no)
+       AC_MSG_ERROR([appstreamcli must have compose support enabled and installed])
+      ])
 
 LT_PREREQ([2.2.6])
 LT_INIT([disable-static])
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -168,7 +168,7 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>appstream-compose</option> (boolean)</term>
-                    <listitem><para>Run <command>appstreamcli-compose</command> during cleanup phase. Defaults to true.</para></listitem>
+                    <listitem><para>Run <command>appstreamcli compose</command> during cleanup phase. Defaults to true.</para></listitem>
                 </varlistentry>
                 <varlistentry>
                     <term><option>sdk-extensions</option> (array of strings)</term>
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -168,7 +168,7 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>appstream-compose</option> (boolean)</term>
-                    <listitem><para>Run appstream-compose during cleanup phase. Defaults to true.</para></listitem>
+                    <listitem><para>Run <command>appstreamcli-compose</command> during cleanup phase. Defaults to true.</para></listitem>
                 </varlistentry>
                 <varlistentry>
                     <term><option>sdk-extensions</option> (array of strings)</term>
--- a/meson.build
+++ b/meson.build
@@ -40,6 +40,10 @@ endforeach
 # The debugedit program is a hard dependency
 debugedit = find_program('debugedit', version: '>= 5.0')
 
+# Require appstream with compose plugin installed
+appstreamcli = find_program('appstreamcli', version: '>= 0.15.0')
+appstreamcli_compose = run_command(appstreamcli, ['--help'], check: true)
+
 subdir('src')
 subdir('doc')
 if get_option('tests')
--- a/meson.build
+++ b/meson.build
@@ -42,7 +42,7 @@ debugedit = find_program('debugedit', version: '>= 5.0')
 
 # Require appstream with compose plugin installed
 appstreamcli = find_program('appstreamcli', version: '>= 0.15.0')
-appstreamcli_compose = run_command(appstreamcli, ['--help'], check: true)
+appstreamcli_compose = run_command(appstreamcli, ['compose', '--help'], check: true)
 
 subdir('src')
 subdir('doc')
--- a/src/builder-context.c
+++ b/src/builder-context.c
@@ -81,6 +81,8 @@ struct BuilderContext
   gboolean        have_rofiles;
   gboolean        run_tests;
   gboolean        no_shallow_clone;
+  gboolean        opt_export_only;
+  char           *opt_mirror_screenshots_url;
 
   BuilderSdkConfig *sdk_config;
 };
@@ -342,6 +344,32 @@ builder_context_find_in_sources_dirs_va (BuilderContext *self,
     }
 
   return NULL;
+}
+
+void
+builder_context_set_opt_export_only (BuilderContext *self,
+                                     gboolean       opt_export_only)
+{
+  self->opt_export_only = opt_export_only;
+}
+
+gboolean
+builder_context_get_opt_export_only (BuilderContext *self)
+{
+  return self->opt_export_only;
+}
+
+void
+builder_context_set_opt_mirror_screenshots_url (BuilderContext *self,
+                                                const char     *opt_mirror_screenshots_url)
+{
+  self->opt_mirror_screenshots_url = g_strdup(opt_mirror_screenshots_url);
+}
+
+const char *
+builder_context_get_opt_mirror_screenshots_url (BuilderContext *self)
+{
+  return self->opt_mirror_screenshots_url;
 }
 
 GFile *
--- a/src/builder-main.c
+++ b/src/builder-main.c
@@ -944,18 +944,8 @@ main (int    argc,
       g_autoptr(GFile) cache = flatpak_build_file (builder_context_get_state_dir (build_context), "screenshots-cache", NULL);
       g_autoptr(GFile) screenshots = flatpak_build_file (app_dir, "screenshots", NULL);
       g_autoptr(GFile) screenshots_sub = flatpak_build_file (screenshots, screenshot_subdir, NULL);
-      g_autofree char *fs_app_dir = g_strdup_printf ("--filesystem=%s", flatpak_file_get_path_cached (app_dir));
-      g_autofree char *fs_cache = g_strdup_printf ("--filesystem=%s", flatpak_file_get_path_cached (cache));
       const char *argv[] = {
-        "flatpak",
-        "build",
-        "--die-with-parent",
-        "--nofilesystem=host:reset",
-        fs_app_dir,
-        fs_cache,
-        "--share=network",
-        flatpak_file_get_path_cached (app_dir),
-        "appstreamcli-compose",
+        "appstreamcli", "compose",
         "--media-baseurl", url,
         "--media-dir", flatpak_file_get_path_cached (screenshots_sub),
         flatpak_file_get_path_cached (xml),
@@ -983,12 +973,12 @@ main (int    argc,
               g_printerr ("Error mirroring screenshots: %s\n", error->message);
               return 1;
             }
-          if (!builder_maybe_host_spawnv (NULL,
-                                          NULL,
-                                          0,
-                                          &error,
-                                          argv,
-                                          NULL))
+          if (!flatpak_spawnv (NULL,
+                               NULL,
+                               0,
+                               &error,
+                               argv,
+                               NULL))
             {
               g_printerr ("Error mirroring screenshots: %s\n", error->message);
               return 1;
--- a/src/builder-main.c
+++ b/src/builder-main.c
@@ -955,12 +955,10 @@ main (int    argc,
         fs_cache,
         "--share=network",
         flatpak_file_get_path_cached (app_dir),
-        "appstream-util",
-        "mirror-screenshots",
+        "appstreamcli-compose",
+        "--media-baseurl", url,
+        "--media-dir", flatpak_file_get_path_cached (screenshots_sub),
         flatpak_file_get_path_cached (xml),
-        url,
-        flatpak_file_get_path_cached (cache),
-        flatpak_file_get_path_cached (screenshots_sub),
         NULL
       };
 
--- a/src/builder-main.c
+++ b/src/builder-main.c
@@ -528,6 +528,8 @@ main (int    argc,
   builder_context_set_jobs (build_context, opt_jobs);
   builder_context_set_rebuild_on_sdk_change (build_context, opt_rebuild_on_sdk_change);
   builder_context_set_bundle_sources (build_context, opt_bundle_sources);
+  builder_context_set_opt_export_only (build_context, opt_export_only);
+  builder_context_set_opt_mirror_screenshots_url (build_context, opt_mirror_screenshots_url);
 
   git_init_email ();
 
@@ -933,60 +935,6 @@ main (int    argc,
 
   if (!opt_require_changes && !opt_export_only)
     builder_cache_ensure_checkout (cache);
-
-  if (opt_mirror_screenshots_url && !opt_export_only)
-    {
-      g_autofree char *screenshot_subdir = g_strdup_printf ("%s-%s", builder_manifest_get_id (manifest),
-                                                            builder_manifest_get_branch (manifest, build_context));
-      g_autofree char *url = g_build_filename (opt_mirror_screenshots_url, screenshot_subdir, NULL);
-      g_autofree char *xml_relpath = g_strdup_printf ("files/share/app-info/xmls/%s.xml.gz", builder_manifest_get_id (manifest));
-      g_autoptr(GFile) xml = g_file_resolve_relative_path (app_dir, xml_relpath);
-      g_autoptr(GFile) cache = flatpak_build_file (builder_context_get_state_dir (build_context), "screenshots-cache", NULL);
-      g_autoptr(GFile) screenshots = flatpak_build_file (app_dir, "screenshots", NULL);
-      g_autoptr(GFile) screenshots_sub = flatpak_build_file (screenshots, screenshot_subdir, NULL);
-      const char *argv[] = {
-        "appstreamcli", "compose",
-        "--media-baseurl", url,
-        "--media-dir", flatpak_file_get_path_cached (screenshots_sub),
-        flatpak_file_get_path_cached (xml),
-        NULL
-      };
-
-      g_print ("Mirroring screenshots from appdata\n");
-      builder_set_term_title (_("Mirroring screenshots"));
-
-      if (!flatpak_mkdir_p (screenshots, NULL, &error))
-        {
-          g_printerr ("Error creating screenshot dir: %s\n", error->message);
-          return 1;
-        }
-
-      if (g_file_query_exists (xml, NULL))
-        {
-          if (!flatpak_mkdir_p (cache, NULL, &error))
-            {
-              g_printerr ("Error creating screenshot cache dir: %s\n", error->message);
-              return 1;
-            }
-          if (!flatpak_break_hardlink (xml, &error))
-            {
-              g_printerr ("Error mirroring screenshots: %s\n", error->message);
-              return 1;
-            }
-          if (!flatpak_spawnv (NULL,
-                               NULL,
-                               0,
-                               &error,
-                               argv,
-                               NULL))
-            {
-              g_printerr ("Error mirroring screenshots: %s\n", error->message);
-              return 1;
-            }
-        }
-
-      g_print ("Saved screenshots in %s\n", flatpak_file_get_path_cached (screenshots));
-    }
 
   if (!opt_build_only &&
       (opt_repo || opt_install) &&
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -2314,9 +2314,9 @@ cmpstringp (const void *p1, const void *p2)
 }
 
 static gboolean
-appstream_compose (GFile   *app_dir,
-                   GError **error,
-                   ...)
+appstreamcli_compose (GFile   *app_dir,
+                      GError **error,
+                      ...)
 {
   g_autoptr(GPtrArray) args = NULL;
   const gchar *arg;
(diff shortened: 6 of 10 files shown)


TingPing:
For CI to work I believe you need to install the `appstream-compose` package.

configure.ac
```diff
@@ -40,7 +41,22 @@ else
...
+   AX_COMPARE_VERSION([$APPSTREAMCLI_REQS],[gt],[$APPSTREAMCLI_VERSION],
+                      [AC_MSG_ERROR([You need at least version $APPSTREAMCLI_REQS of appstreamcli, your version is $APPSTREAMCLI_VERSION])])
+fi
 
+AC_MSG_CHECKING([whether appstreamcli has compose support])
+AS_IF([appstreamcli compose --help >/dev/null 2>/dev/null],
+      [AC_MSG_RESULT(yes)],
+      [
+       AC_MSG_RESULT(no)
+       AC_MSG_ERROR([appstreamcli must have compose support enabled])
```
TingPing, line 57:
```suggestion
       AC_MSG_ERROR([appstreamcli must have compose support enabled and installed])
```

I think this is more clear because its a separate package in many cases.

src/builder-context.c
```diff
@@ -344,6 +346,28 @@ builder_context_find_in_sources_dirs_va (BuilderContext *self,
...
+builder_context_set_opt_mirror_screenshots_url (BuilderContext *self,
+                                                     char *opt_mirror_screenshots_url){
+  self->opt_mirror_screenshots_url=opt_mirror_screenshots_url;
+}
+
+char *
+builder_context_get_opt_mirror_screenshots_url (BuilderContext *self){
+  return self->opt_mirror_screenshots_url;
+}
+
```
TingPing, line 370: Nitpick about weird formatting throughout here. `{` should be on new lines. params should be aligned.

src/builder-manifest.c
```diff
@@ -2773,15 +2767,44 @@ builder_manifest_cleanup (BuilderManifest *self,
...
-                                  basename_arg,
-                                  self->id,
-                                  NULL))
+          g_autofree char *components_arg = g_strdup_printf ("--components=%s", self->id);
+          g_autofree char *app_root_path = g_file_get_path (app_root);
+          g_autofree char *result_root_arg = g_strdup_printf ("--result-root=%s", app_root_path);
+          g_autofree char *opt_mirror_screenshots_url = builder_context_get_opt_mirror_screenshots_url(context);
+          gboolean opt_export_only = builder_context_get_opt_export_only(context);
+
+          if ( opt_mirror_screenshots_url && !opt_export_only) {
```
TingPing, line 2776: Another formatting nitpick. This uses GNU style indenting for `{`.

TingPing:
NOTE: This is blocked by https://github.com/flatpak/flatpak/pull/5277 making into a release.

JakobDev:
> NOTE: This is blocked by flatpak/flatpak#5282 making it into a release.

It looks like this PR only adds a new Path to search for the AppSteram catalogue. Saving the catalogue under thee old location, until a new Release happens, should solve this.

Lunarequest:
it's not really possible without breaking everything

barthalion:
As mentioned under flatpak's PR, I think we should keep using the `app-info` location.

Lunarequest:
in theory you will no longer need https://github.com/flatpak/flatpak/pull/5282 since this is one-to-one with the old output and the linked pr is pure compact work

libglnx-config.h
```diff
@@ -0,0 +1 @@
+#include "config.h"
```
refi64, line 1: This should be `.gitignore`'d like in the Flatpak repo.

src/builder-context.c
```diff
@@ -81,6 +81,8 @@ struct BuilderContext
   gboolean        have_rofiles;
   gboolean        run_tests;
   gboolean        no_shallow_clone;
+  gboolean        opt_export_only;
+  char            *opt_mirror_screenshots_url;
```
refi64, line 85: style nit:
```suggestion
  char           *opt_mirror_screenshots_url;
```

src/builder-context.c
```diff
@@ -344,6 +346,32 @@ builder_context_find_in_sources_dirs_va (BuilderContext *self,
...
+builder_context_get_opt_export_only (BuilderContext *self)
+{
+  return self->opt_export_only;
+}
+
+void
+builder_context_set_opt_mirror_screenshots_url (BuilderContext *self,
+                                                     char *opt_mirror_screenshots_url)
+{
+  self->opt_mirror_screenshots_url=opt_mirror_screenshots_url;
```
refi64, line 366: This needs to be freed in `builder_context_finalize`, and the function should probably take a `const char*` and `g_strdup` it instead of taking ownership? Also, style nit:

```suggestion
  self->opt_mirror_screenshots_url = opt_mirror_screenshots_url;
```

src/builder-context.c
```diff
@@ -344,6 +346,32 @@ builder_context_find_in_sources_dirs_va (BuilderContext *self,
   return NULL;
 }
 
+void 
+builder_context_set_opt_export_only (BuilderContext *self,
+                                          gboolean       opt_export_only)
```
refi64, line 351: Some weird things going on with this second argument indentation & `builder_context_set_opt_mirror_screenshots_url`...

src/builder-main.c
```diff
@@ -528,6 +528,8 @@ main (int    argc,
   builder_context_set_jobs (build_context, opt_jobs);
   builder_context_set_rebuild_on_sdk_change (build_context, opt_rebuild_on_sdk_change);
   builder_context_set_bundle_sources (build_context, opt_bundle_sources);
+  builder_context_set_opt_export_only(build_context, opt_export_only);
+  builder_context_set_opt_mirror_screenshots_url(build_context, opt_mirror_screenshots_url);
```
refi64, line 532: style nit: these parens on the function calls & the others here should have a space after the function name

src/builder-context.c
```diff
@@ -344,6 +346,32 @@ builder_context_find_in_sources_dirs_va (BuilderContext *self,
...
+}
+
+void
+builder_context_set_opt_mirror_screenshots_url (BuilderContext *self,
+                                                     char *opt_mirror_screenshots_url)
+{
+  self->opt_mirror_screenshots_url=opt_mirror_screenshots_url;
+}
+
+char *
```
refi64, line 369: Should be `const char *`, since this class still owns the string.

src/builder-manifest.c
```diff
@@ -2773,17 +2767,56 @@ builder_manifest_cleanup (BuilderManifest *self,
...
-                                  NULL))
+          g_autofree char *origin = g_strdup_printf ("--origin=%s",builder_manifest_get_id (self));
+          g_autofree char *components_arg = g_strdup_printf ("--components=%s", self->id);
+          g_autofree char *app_root_path = g_file_get_path (app_root);
+          g_autofree char *result_root_arg = g_strdup_printf ("--result-root=%s", app_root_path);
+          g_autoptr(GFile) xml_dir = flatpak_build_file (app_root, "share/app-info/xmls", NULL);
+          g_autoptr(GFile) icon_out = flatpak_build_file (app_root, "share/app-info/icons/flatpak", NULL);
+          g_autofree char *data_dir = g_strdup_printf("--data-dir=%s",flatpak_file_get_path_cached(xml_dir));
+          g_autofree char *icon_dir = g_strdup_printf("--icons-dir=%s",flatpak_file_get_path_cached(icon_out));
+          g_autofree char *opt_mirror_screenshots_url = builder_context_get_opt_mirror_screenshots_url(context);
```
refi64, line 2778: This shouldn't be `g_autofree` and just `const char *` instead.

src/builder-manifest.c
```diff
@@ -2773,17 +2767,56 @@ builder_manifest_cleanup (BuilderManifest *self,
...
         {
-          g_autofree char *basename_arg = g_strdup_printf ("--basename=%s", self->id);
-          g_print ("Running appstream-compose\n");
-          if (!appstream_compose (app_dir, error,
-                                  self->build_runtime ?  "--prefix=/usr" : "--prefix=/app",
-                                  "--origin=flatpak",
-                                  basename_arg,
-                                  self->id,
-                                  NULL))
+          g_autofree char *origin = g_strdup_printf ("--origin=%s",builder_manifest_get_id (self));
```
refi64, line 2770: style nit:
```suggestion
          g_autofree char *origin = g_strdup_printf ("--origin=%s", builder_manifest_get_id (self));
```

(same for elsewhere in this file)

src/builder-manifest.c
```diff
@@ -2773,17 +2767,56 @@ builder_manifest_cleanup (BuilderManifest *self,
...
+          g_autofree char *app_root_path = g_file_get_path (app_root);
+          g_autofree char *result_root_arg = g_strdup_printf ("--result-root=%s", app_root_path);
+          g_autoptr(GFile) xml_dir = flatpak_build_file (app_root, "share/app-info/xmls", NULL);
+          g_autoptr(GFile) icon_out = flatpak_build_file (app_root, "share/app-info/icons/flatpak", NULL);
+          g_autofree char *data_dir = g_strdup_printf("--data-dir=%s",flatpak_file_get_path_cached(xml_dir));
+          g_autofree char *icon_dir = g_strdup_printf("--icons-dir=%s",flatpak_file_get_path_cached(icon_out));
+          g_autofree char *opt_mirror_screenshots_url = builder_context_get_opt_mirror_screenshots_url(context);
+          gboolean opt_export_only = builder_context_get_opt_export_only(context);
+
+          if ( opt_mirror_screenshots_url && !opt_export_only)
```
refi64, line 2781:
```suggestion
          if (opt_mirror_screenshots_url && !opt_export_only) 
```

src/builder-manifest.c
```diff
@@ -2773,17 +2767,56 @@ builder_manifest_cleanup (BuilderManifest *self,
...
+                                      data_dir,
+                                      icon_dir,
+                                      components_arg,
+                                      app_root_path,
+                                      NULL))
+              return FALSE;
+          }
         }
-
+
```
refi64, line 2819: Evil trailing whitespace snuck in:

```suggestion

```

src/builder-manifest.c
```diff
@@ -2773,17 +2767,56 @@ builder_manifest_cleanup (BuilderManifest *self,
...
-          g_print ("Running appstream-compose\n");
-          if (!appstream_compose (app_dir, error,
-                                  self->build_runtime ?  "--prefix=/usr" : "--prefix=/app",
-                                  "--origin=flatpak",
-                                  basename_arg,
-                                  self->id,
-                                  NULL))
+          g_autofree char *origin = g_strdup_printf ("--origin=%s",builder_manifest_get_id (self));
+          g_autofree char *components_arg = g_strdup_printf ("--components=%s", self->id);
+          g_autofree char *app_root_path = g_file_get_path (app_root);
```
refi64, line 2772: I believe this can use `flatpak_file_get_path_cached` instead.

src/builder-manifest.c
```diff
@@ -2773,17 +2767,56 @@ builder_manifest_cleanup (BuilderManifest *self,
...
+                                      components_arg,
+                                      app_root_path,
+                                      NULL))
             return FALSE;
+          }
+          else 
+          {
+            g_print ("Running appstreamcli compose\n");
+            if (!appstreamcli_compose (error,
+                                      "--prefix=/",
```
refi64, line 2808: style nit: here & above the indentation is a bit off

src/builder-manifest.c
```diff
@@ -2773,17 +2767,56 @@ builder_manifest_cleanup (BuilderManifest *self,
...
+          g_autofree char *result_root_arg = g_strdup_printf ("--result-root=%s", app_root_path);
+          g_autoptr(GFile) xml_dir = flatpak_build_file (app_root, "share/app-info/xmls", NULL);
+          g_autoptr(GFile) icon_out = flatpak_build_file (app_root, "share/app-info/icons/flatpak", NULL);
+          g_autofree char *data_dir = g_strdup_printf("--data-dir=%s",flatpak_file_get_path_cached(xml_dir));
+          g_autofree char *icon_dir = g_strdup_printf("--icons-dir=%s",flatpak_file_get_path_cached(icon_out));
+          g_autofree char *opt_mirror_screenshots_url = builder_context_get_opt_mirror_screenshots_url(context);
+          gboolean opt_export_only = builder_context_get_opt_export_only(context);
+
+          if ( opt_mirror_screenshots_url && !opt_export_only) 
+          {
```
refi64, line 2782: style nit: braces should be one level indented

src/builder-manifest.c
refi64: Hmm based on glancing at the previous behavior, I think *both* clauses of this should be behind `if (!opt_export_only)`, and then this becomes `if (opt_mirror_screenshots_url)`? Since otherwise, leaving off opt_export_only will still result in running `appstreamcli`, which I don't *think* is what we want?

src/builder-manifest.c
```diff
@@ -2773,17 +2767,56 @@ builder_manifest_cleanup (BuilderManifest *self,
...
+            g_autofree char *screenshot_subdir = g_strdup_printf ("%s-%s", builder_manifest_get_id (self), builder_manifest_get_branch (self, context));
+            g_autofree char *url = g_build_filename (opt_mirror_screenshots_url, screenshot_subdir, NULL);
+            g_autoptr(GFile) screenshots = flatpak_build_file (app_root, "screenshots", NULL);
+            g_autofree char *arg_base_url = g_strdup_printf("--media-baseurl=%s",url);
+            g_autofree char *arg_media_dir =  g_strdup_printf ("--media-dir=%s",  flatpak_file_get_path_cached (screenshots));
```
refi64, line 2787: style nit:
```suggestion
            g_autofree char *arg_media_dir = g_strdup_printf ("--media-dir=%s", flatpak_file_get_path_cached (screenshots));
```

configure.ac
```diff
@@ -40,7 +41,22 @@ else
...
+AC_CHECK_PROG([APPSTREAMCLI], [appstreamcli], [appstreamcli], [false])
+if test "x$APPSTREAMCLI" = xfalse; then
+   AC_MSG_ERROR([You need appstreamcli installed])
+   APPSTREAMCLI_VERSION=`$APPSTREAMCLI --version | sed 's,.*\ \([0-9]*\.[0-9]*\.[0-9]*\)$,\1,'`
+   AX_COMPARE_VERSION([$APPSTREAMCLI_REQS],[gt],[$APPSTREAMCLI_VERSION],
+                      [AC_MSG_ERROR([You need at least version $APPSTREAMCLI_REQS of appstreamcli, your version is $APPSTREAMCLI_VERSION])])
+fi
 
+AC_MSG_CHECKING([whether appstreamcli has compose support])
+AS_IF([appstreamcli compose --help >/dev/null 2>/dev/null],
```
refi64, line 53:
```suggestion
AS_IF([appstreamcli compose --help >/dev/null 2>&1],
```

src/builder-manifest.c
refi64: UPDATE: no I just can't read code, this checks out, it's just weird semantics to begin with.

src/builder-manifest.c
Lunarequest: evil trailing whitespace has been slain

Lunarequest:
> > in theory you will no longer need [flatpak/flatpak#5282](https://github.com/flatpak/flatpak/pull/5282) since this is one-to-one with the old output and the linked pr is pure compact work
> 
> Might still be useful to have that patch, so we could switch eventually, many years in the future (so, have it as a future-compatibility thing).

I doubt we will ever switch to `swcatalog` unless someone backports the pr to all versions of flatpak currently in the wild. It should be possible but I doubt the benefits would actually be worth doing this. (this is not including all the infra and 3rd part programs that may depend on the old paths)

darkdragon-001:
So this is the final missing piece towards support of `<requires>`/`<recommends>` elements?

This is blocking software catalogs like [Apps for GNOME](https://gitlab.gnome.org/World/apps-for-gnome/-/issues/68) to rely on AppStream data via [Flathub](https://github.com/flathub/flathub/issues/2439).

ximion:
@darkdragon-001 As far as I know yes, this is the last piece.

ximion:
Only if you don't use flatpak-builder - Flatpak itself handles all data just fine since it has been using libappstream for a while now, but when building a Flatpak, appstream-glib will filter some stuff out or throw warnings about unknown entries. So you can not be sure that the data that is generated is actually complete right now.

darkdragon-001:
@refi64 @TingPing @barthalion could someone of you please review and merge if everything is fine?

barthalion:
This is a potentially breaking change for Flathub, so it's not going to be merged and deployed until I have time to test it in about 2 weeks.

JakobDev:
@barthalion Any News on this?

barthalion:
As it is, I consider it good to be merged. I will clean up commits history later today, and tag a new unstable release after the merge.

JakobDev:
When will this be on Flathub? Most of my Apps use newer features of AppStream such as `<controls>` that are not supported by appstream-glib. And does the Apps needs a rebuild after that, so the Tags that are only supported by appstreamcli will work?

barthalion:
There are changes needed across flat-manager, buildbot, nginx configuration and the backend, and it will need to be rolled out together so it all doesn't fall apart. Might happen this week.

Yes, the apps using previously unsupported tags will need to be rebuilt.

meson.build
```diff
@@ -40,6 +40,10 @@ endforeach
 # The debugedit program is a hard dependency
 debugedit = find_program('debugedit', version: '>= 5.0')
 
+# Require appstream with compose plugin installed
+appstreamcli = find_program('appstreamcli', version: '>= 0.15.0')
+appstreamcli_compose = run_command(appstreamcli, ['--help'], check: true)
```
wjt, line 45: This doesn't check whether the compose plugin is installed. I think this should be changed to match the `configure.ac` check thusly:

```suggestion
appstreamcli_compose = run_command(appstreamcli, ['compose', '--help'], check: true)
```

doc/flatpak-manifest.xml
```diff
@@ -168,7 +168,7 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>appstream-compose</option> (boolean)</term>
-                    <listitem><para>Run appstream-compose during cleanup phase. Defaults to true.</para></listitem>
+                    <listitem><para>Run <command>appstreamcli-compose</command> during cleanup phase. Defaults to true.</para></listitem>
```
wjt, line 171: I think it's a subcommand?:

```suggestion
                    <listitem><para>Run <command>appstreamcli compose</command> during cleanup phase. Defaults to true.</para></listitem>
```

wjt (commented):
I only skim-read this but I spotted a couple of minor points

meson.build
Lunarequest: No this should be the other way around we specifically need compose

barthalion: Good catch, thanks

wjt: Yes, we specifically need compose, and the Meson check currently doesn't check for it.

ximion: Hmm, maybe the CLI tools should also ship pkg-config files, to make version and presence checking easier...

Lunarequest: it does ship pkg-config files

Lunarequest: image
on opensuse it ships under `AppStream-compose-devel`

ximion: > both appstream and appstream compose ship pkg-config files

Yes, but for the library development files, not for the command-line utility. You can have the libraries without CLI...

Lunarequest: > > both appstream and appstream compose ship pkg-config files
> 
> Yes, but for the library development files, not for the command-line utility. You can have the libraries without CLI...

given we actually call [`appstreamcli compose`](https://github.com/flatpak/flatpak-builder/blob/8c98ddb8f5b2ee3a63265b984f4c46e434aab444/meson.build#L45) I think @barthalion is right we do need to check for the `appstreamcli compose` command vs the libs

Lunarequest:
is there any update? its been 2 weeks with no word on this, is there something holding this up that needs to be worked on?

kparal:
This is not yet deployed to production, right? I just tested adding `<code>` and `<em>` to app description, and it still shows up literally in gnome-software. Also [here](https://github.com/organicmaps/organicmaps/issues/4588#issuecomment-1578412267) someone claims it works, but on the [linked web](https://flathub.org/apps/app.organicmaps.desktop), I again see literal `<em>` tags.

ximion:
> We had to revert it due to various problems.

Which problems? This one https://github.com/flatpak/flatpak/issues/5434? Or were there more? Apparently the issue doesn't happen anymore (potentially due to the revert @ Flathub) which actually makes testing the fixed appstream-glib versions on stable distributions like Ubuntu a bit harder, since the original cause for the SRU has suddenly vanished.

razzeee:
I think that was one of the most annoying ones.

https://github.com/flathub/flathub/issues/4222 played a role too, I think.

@barthalion might have a clearer picture

jbicha:
Hi, I fixed appstream-glib for Ubuntu 22.04 LTS and was in the process of verifying the fix for Ubuntu 20.04 LTS. Do y'all expect to later re-enable the code/em feature on Flathub? If the answer is no, then I might not be able to convince the Ubuntu Stable Release Update team to land the fix for Ubuntu 20.04 LTS because we try not to push unnecessary updates to our users.

ximion:
> [flathub/flathub#4222](https://github.com/flathub/flathub/issues/4222) played a role too, I think.

3rd-party AppStream parsers are such a pain... It's actually a really complex specification now with many corner cases to consider.
For this particular thing, I think libappstream has a knob to not use `media_baseurl` when generating data, it just wasn't used at all so far... I'll have a look later :-) (that Rust thing should be fixed of course, but maybe improving `compose` with a few more options is also useful).

But in any case, that's good to know, with that information maybe the Ubuntu SRU team can be convinced to let the appstream-glib update in to to fix https://github.com/flatpak/flatpak/issues/5434, even though the test case for it has vanished and can't be verified anymore. Let's see :-)

barthalion:
> Hi, I fixed appstream-glib for Ubuntu 22.04 LTS and was in the process of verifying the fix for Ubuntu 20.04 LTS. Do y'all expect to later re-enable the code/em feature on Flathub? If the answer is no, then I might not be able to convince the Ubuntu Stable Release Update team to land the fix for Ubuntu 20.04 LTS because we try not to push unnecessary updates to our users.

Hi Jeremy, the plan is to re-enable it. I pulled it out exactly due to broken Ubuntu 20.04/22.04, so we're still interested in seeing it fixed via SRU. I can provide a testing repo with libappstream-generated files if it helps with anything.

> For this particular thing, I think libappstream has a knob to not use media_baseurl when generating data, it just wasn't used at all so far... I'll have a look later :-)

Is it actually exposed via CLI?

Lunarequest:
as far as i can tell, if it is its not documented in the man page

barthalion:
I'm looking into setting up a regular repo somewhere, but in the meantime, this is the file with code/em tags included: [appstream.xml.gz](https://github.com/flatpak/flatpak-builder/files/11902930/appstream.xml.gz)

It can be moved to `~/.local/share/flatpak/appstream/flathub/x86_64/active` and then decompressed to overwrite `appstream.xml` too. As long as you (or something else like GNOME Software) don't run `flatpak update`, this will do. Although I can imagine flatpak can be a little aggressive about modifications like this so might be better to issue the search command offline.

jbicha:
That directory does not exist on my Ubuntu 20.04 LTS system. I disconnected from the Internet, copied the file there (both compressed and uncompressed) and the `flatpak search` command still doesn't reproduce the failure.

smcv:
@jbicha: It'd be `/var/lib/flatpak/appstream/flathub/x86_64/active` if you only have Flathub added as a remote in the system installation, which is [the default/recommended setup](https://flatpak.org/setup/Ubuntu). Is that any help?

barthalion:
This is somewhat big, but I verified it fails "properly": https://dl.flathub.org/repo/assets/repo.tar.zst

Steps to reproduce in a Docker container:

```
docker run --rm -it ubuntu:20.04 bash
apt-get update
apt-get install -y flatpak wget zstd
wget https://dl.flathub.org/repo/assets/repo.tar.zst
tar xf repo.tar.zst
flatpak remote-add --no-gpg-verify test file:///repo
flatpak update --appstream
flatpak search map
```

ximion:
Can we flip on the AppStream/appstreamcli support in Flathub/flatpak-builder again?
@barthalion, have you tried the `--no-partial-urls` flag on `appstreamcli compose`? That should make all URLs complete URLs, so even very old clients will have no issue reading them.

ximion:
Can we flip on the AppStream/appstreamcli support in Flathub/flatpak-builder again?
@barthalion, have you tried the `--no-partial-urls` flag on `appstreamcli compose`? That should make all URLs complete URLs, so even very old clients will have no issue reading them.

@razzeee Are there any remaining issues you know of? Because using the unmaintained old AppStream tools is becoming a massive problem on its own, and I now see people running into weird bugs and frustrations multiple times a month - and they report bugs against upstream AppStream, which I can't help them with because the issue lies in the old tooling in Flatpak.

razzeee:
Yeah, I don't think we know of anything that needs solving from your side of things. Basically waiting for barth to find some time to look at it again and switch us over again.

PR #520, opened 2023-02-12
Add argument to install debug extension

ColinKinloch:
Adds an argument `--install-debug` to install `package-id.Debug`


TingPing:
> Ideally though I think --install (at least for apps) should handle by default both .Locale, .Debug and any app specific extensions specified in the app manifest.

I think I agree. It seems unlikely many would want to test without these.

PR #525, opened 2023-03-13
Improve grammar in doc/flatpak-manifest.xml

jwakely:
I'm not sure what the sentence "Extension define extension points in the app/runtime that can be implemented by extensions, supplying extra files which are available during runtime" is really trying to say (it uses the word extension too many times!) but I think this is better.

I'm also not sure exactly what the `no-debuginfo` description is saying. Please check my changes make sense.


doc/flatpak-manifest.xml
```diff
@@ -132,11 +132,11 @@
                     <term><option>command</option> (string)</term>
                     <listitem><para>
                         The filename or path to the main binary of the application. Note that this
-                        is really just a single file, not a commandline. If you want to pass arguments,
+                        must be just a single file, not a commandline. If you want to pass arguments,
                         install a shell script wrapper and use that as the command.
                     </para><para>
                         Also note that the command is used when the application is run via
-                        <command>flatpak run</command>, and does not affect what gets executed when
+                        <command>flatpak run</command>, and does not necessarily affect what gets executed when
```
jwakely, line 139: I made this change because (if I understand correctly) the `command` will get used by a .desktop file unless it overrides it with `--command`. So it's wrong to say it categorically does not affect desktop files.

Maybe it should mention that it can be overridden by the `--command` command-line option, and say that typically desktop files use that?

jwakely:
> You could probably omit it and say:

Yes, the same thought had occurred to me. I'll push an update to the branch.

doc/flatpak-manifest.xml
jwakely: If a desktop file has `Exec=flatpak run com.example.App` then it will be affected by this.

If a desktop file has `Exec=flatpak run --command=app-launcher.sh com.example.App` then it won't be affected by this.

So unless I'm misunderstanding something, a desktop file *might* be affected by this, but it depends what the desktop's Exec command does. So it's wrong to say it categorically doesn't use it. And if it's wrong because desktop files can do anything they want, why mention a desktop file at all?

Just say that this is the default command for `flatpak run ...` but that it's possible to run other commands from the flatpak application.

TingPing: A desktop file cannot have `flatpak run` without a command.

jwakely: > A desktop file cannot have `flatpak run` without a command.

Why not? Is that just a convention? I can create a desktop file with `Exec=/usr/bin/flatpak run com.example.App` and `desktop-file-validate` doesn't complain about it. https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html doesn't say I can't do that.

https://docs.flatpak.org/en/latest/freedesktop-quick-reference.html#desktop-files even gives an example of such a desktop file!

> This way, the generated Exec line looks like this:
>  
> `Exec=/usr/bin/flatpak run --branch=stable --arch=x86_64 org.gnome.Dictionary`

I'll revert this part of the commit, but *something* is wrong in the docs somewhere. Either this is wrong, or the example above is wrong.

TingPing: flatpak-builder makes the desktop file and controls its content. This isn’t about theoretical files but real world usage.

TingPing: Wow, I did not know you can do ‘Exec=‘. That is a very odd thing to do.

doc/flatpak-manifest.xml
```diff
@@ -411,24 +411,24 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>no-debuginfo</option> (boolean)</term>
-                    <listitem><para>By default (if strip is not true) flatpak-builder extracts all debug info in ELF files to a separate files
-                    and puts this in an extension. If you want to disable this, set no-debuginfo to true.</para></listitem>
+                    <listitem><para>By default (if strip is not true) flatpak-builder extracts all debug info in ELF files to separate files
+                    and puts them in an extension. If you want to disable this, set no-debuginfo to true.</para></listitem>
```
hfiguiere, line 415:
```suggestion
                    and puts them into an extension. If you want to disable this, set no-debuginfo to true.</para></listitem>
```

doc/flatpak-manifest.xml
```diff
@@ -132,7 +132,7 @@
                     <term><option>command</option> (string)</term>
                     <listitem><para>
                         The filename or path to the main binary of the application. Note that this
-                        is really just a single file, not a commandline. If you want to pass arguments,
+                        must be just a single file, not a commandline. If you want to pass arguments,
```
hfiguiere, line 135:
```suggestion
                        must be just a single command name, not a commandline, nor path. If you want to pass arguments,
                        or run from a location not in PATH,
```

doc/flatpak-manifest.xml
```diff
@@ -377,7 +377,7 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>secret-env</option> (array of strings)</term>
-                    <listitem><para>This is a array defining which host environment variables is transfered to build-commands or post-install environment.</para></listitem>
+                    <listitem><para>This is a array defining which host environment variables are transfered to the build-commands or post-install environment.</para></listitem>
```
hfiguiere, line 380:
```suggestion
                    <listitem><para>This is an array defining which host environment variables are transfered to the build-commands or post-install environment.</para></listitem>
```

doc/flatpak-manifest.xml
bbhtt:
```suggestion
                        must be a single executable, not a commandline or an absolute path. If you want to pass arguments,
```

bbhtt: went with https://github.com/flatpak/flatpak-builder/pull/525#discussion_r2438005828 instead

doc/flatpak-manifest.xml
bbhtt: applied

doc/flatpak-manifest.xml
bbhtt: applied

PR #526, opened 2023-03-13
Document that config-opts is used for CMake

jwakely:
It's not clear from the docs that `config-opts` is used for the cmake buildsystem as well as autotools.

### DIFF
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -389,7 +389,7 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>config-opts</option> (array of strings)</term>
-                    <listitem><para>This is an array containing extra options to pass to configure or cmake.</para></listitem>
+                    <listitem><para>This is an array containing extra options passed to the build system during configuration.</para></listitem>
                 </varlistentry>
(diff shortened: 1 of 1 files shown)


doc/flatpak-manifest.xml
```diff
@@ -389,7 +389,7 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>config-opts</option> (array of strings)</term>
-                    <listitem><para>This is an array containing extra options to pass to configure.</para></listitem>
+                    <listitem><para>This is an array containing extra options to pass to configure or cmake.</para></listitem>
```
TingPing, line 392:
```suggestion
                    <listitem><para>This is an array containing extra options passed to the build system during configuration.</para></listitem>
```

TingPing: This applies to all build systems, is this still clear to you?

jwakely: Thanks - the docs made it looks like it only applied if running `configure`, and I found by trial and error that it applies to cmake too. Your new phrasing avoids the implication that it only applies if there's a `configure` script, thanks.

Issue #528, opened 2023-03-17
Meson port doesn't support installed-tests


TingPing:
The new meson port doesn't support installed-tests. Meson does not provide any built-in way to do this so it must be done by hand.

See:

https://github.com/flatpak/flatpak-builder/blob/e6ff3de4b15e1f87f1a7d960eed59f0295ef1fb3/tests/Makefile.am.inc
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/359/diffs
https://github.com/mesonbuild/meson/issues/186

smcv:
This is not usually very difficult, and it's a blocker for me packaging 1.3.2 in Debian experimental, so I'll try to propose a PR for this.

PR #532, opened 2023-03-18
meson: Reinstate parallel tests, add as-installed testing

smcv:
* tests: Run each test in a new, empty directory
    
    The test framework in libtest.sh expects this, and it's what the
    Autotools build system did.

* tests: Run in parallel
    
    Now that each test has its own working directory, hopefully we can run
    them in parallel.

* tests: Add GNOME-style installed tests to Meson build
    
    This was supported with Autotools, but missing from Meson until now.
    
    Resolves: https://github.com/flatpak/flatpak-builder/issues/528

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -87,6 +87,8 @@ EXTRA_DIST += 			\
 	doc/meson.build		\
 	src/meson.build		\
 	tests/meson.build \
+	tests/test-keyring/meson.build \
+	tests/test-keyring2/meson.build \
 	subprojects/libglnx.wrap \
 	subprojects/libglnx/meson.build \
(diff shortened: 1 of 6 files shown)


smcv:
After doing an Autotools build + install with `DESTDIR=.../autotools.inst` and a Meson build + install with `DESTDIR=.../meson.inst`, both with a distro-style `--prefix` etc., and both enabling installed-tests:

```
$ diff -ru ~/tmp/build/flatpak-builder/*.inst
Binary files .../autotools.inst/usr/bin/flatpak-builder and .../meson.inst/usr/bin/flatpak-builder differ
Only in .../autotools.inst/usr/libexec: flatpak-builder-debugedit
```

Issue #529, opened 2023-03-17
Fix TAP output of tests under Meson


TingPing:
The existing tests were loosely based on TAP however there were cases of invalid output, mostly `flatpak-builder` using `g_print()` instead of `g_printerr()` when something fails.

PR #536, opened 2023-03-21
Use TAP for Meson tests

smcv:
* tests: Send diagnostic output from Flatpak etc. to stderr
    
    This avoids interfering with the machine-readable TAP output on stdout,
    which should be interpreted by Meson.
    
* tests: Turn an unstructured "echo" into a TAP diagnostic[

* tests: Tell Meson to parse TAP output
    
    Resolves: https://github.com/flatpak/flatpak-builder/issues/529

---

I haven't really tested `test-builder-python.sh` since my distribution no longer supports Python 2, but I think it's right.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -312,7 +312,7 @@ skip_without_python2 () {
 cleanup () {
     gpg-connect-agent --homedir "${FL_GPG_HOMEDIR}" killagent /bye >&2 || true
     if test -n "${TEST_SKIP_CLEANUP:-}"; then
-        echo "Skipping cleanup of ${TEST_DATA_DIR}"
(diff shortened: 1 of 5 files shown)


smcv:
I'm aware of how TAP works, but flatpak and flatpak-builder are general-purpose semi-interactive commands that output arbitrary human-readable text, not TAP, so their output is not going to change to have a `#` prefix for the benefit of the tests.

Putting them through `| sed -e 's/^/# /'` would make it more difficult to get their exit status, so I'd prefer not to do that either.

The remaining option is to have them write to stderr, which has the benefit that their output interleaves correctly with the `set -x` trace to make it easier to debug failing tests.

TingPing:
I agree, this is the more practical approach.

PR #540, opened 2023-04-08
improve editor experience by adding a `title`

proletarius101:
To avoid displaying the entire schema URL.

https://github.com/SchemaStore/schemastore/blob/master/editor-features.md#title-as-an-expected-object-type


data/flatpak-manifest.schema.json
```diff
@@ -1,6 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema",
   "$comment": "https://docs.flatpak.org/en/latest/flatpak-builder-command-reference.html#flatpak-manifest",
+  "title": "Flatpak",
```
TingPing, line 4:
```suggestion
  "title": "flatpak-builder manifest",
```

TingPing: Seems more clear to what it is.

Issue #439, opened 2021-12-06
Archive-source with "git-init" option may cause step to fail waiting for gpg signing


cobratbq:
This issue is related to https://github.com/flathub/shared-modules/issues/165, where I first reported it.
If gpg signing is enabled by default (global option) then the commit step may never complete without user interaction, due to git waiting for user interaction to complete gpg signing. This seems unnecessary, given the use case.

Just now, I realized that there is a command-line option `--no-gpg-sign` which may be of use. I think it is a matter of adding the above parameter to the commit operation. (Or, alternatively, setting `commit.gpgsign` `false` config option.) 

https://github.com/flatpak/flatpak-builder/blob/a9cee912be36767ca4065e2060abcbc845f6ef0e/src/builder-source-archive.c#L670

PR #541, opened 2023-04-11
git-am: add --no-gpg-sign to override config

cobratbq:
Patches applied with `git am` require committing. If `git` is configured to require signing for commits, then this will block flatpak-builder. The argument overrides `commit.gpgsign=true`. (`git config --global commit.gpgsign true`)

I have not tested this myself. Is it possible to test this on an existing set-up?


bbhtt:
Seems fine, should probably use `--no-verify` too to skip hooks.

bbhtt:
Fixes https://github.com/flatpak/flatpak-builder/issues/439

bbhtt:
@cobratbq can you please rebase?

PR #546, opened 2023-04-25
builder-module: run autotools tests in parallel

lw64:


lw64:
I am not sure if parallel tests should be also enabled for other build systems.

TingPing:
The same `-j` arguments apply to both cmake and meson.

lw64:
I think Meson is doing it by default though. Should it still be given there too?

TingPing:
Yes, `ninja` defaults to number of threads + 2 or similar.

However I think this is a bugfix, because now it will respect limits you give `flatpak-builder`.

lw64:
Thanks! Is there going to be a release with this fix soon?

PR #551, opened 2023-05-14
builder-git: trim whitespace from submodule path

hfiguiere:
Without this patch https://github.com/nidefawl/daw-deps/blob/master/.gitmodules is not parsed properly causing `tinyspline` to never be checked out because it check for `tinyspline` since there is a whitespace at the end of the `path = `.


hfiguiere:
Without #510 (older version of flaptak-builder), I get an early bail out `Failed to download sources: module bass-studio: Not a gitlink tree: tinyspline ` (notice the space)

PR #559, opened 2023-09-03
builder-module: Use `meson setup` instead of `meson`

ItsJamie9494:
Running just `meson` is deprecated and should be replaced with the explicit `meson setup` ([Reference](https://mesonbuild.com/Commands.html#setup))


smcv:
My only concern about this is: when was the ability to run `meson setup` added, and are there any Flatpak runtimes still supported that are sufficiently old to not have it?

smcv:
> My only concern about this is: when was the ability to run `meson setup` added

Meson 0.49.2 (in Debian 10, released in 2019) already had `meson setup`, so I think that's good enough.

smcv (approved):

Issue #15, opened 2017-08-25
Discuss how to package Rust/Cargo applications (and other non-standard build systems)


alexlarsson:
_From @moosingin3space on February 6, 2017 17:38_

When I read the documentation for `flatpak-builder`, I see that it wants a project to support the Build API. Rust applications that build with Cargo don't fit as nicely into this mold. I'm curious what the expectation is for a non-standard build system. Should the build system be extended to produce Flatpaks itself, or should `flatpak-builder` be extended to support arbitrary build systems, in a similar manner to Arch Linux `PKGBUILD`s?

_Copied from original issue: flatpak/flatpak#536_

alexlarsson:
Oh, the build api just makes it easier to build with f-b, you can build whatever you want if you add a custom makefile. Alternatively we're making that not required by https://github.com/flatpak/flatpak/pull/516

alexlarsson:
_From @TingPing on February 6, 2017 18:49_

The "Build API" should probably not be mentioned in the first place, it is extremely autotools specific and even already `flatpak-builder` supports CMake and Meson.

alexlarsson:
_From @moosingin3space on February 6, 2017 18:59_

So, regarding Rust/Cargo specifically, we need:
- The ability to run `cargo fetch` or `cargo vendor` or something *outside* the sandbox, so all dependencies can be fetched.
- Run `cargo build --release` *inside* the sandbox
- `cp` some files around to fit the structure of a flatpak

Is all this possible with `flatpak-builder`? 

I'm also interested in packaging Servo, which has a more complicated build system. The system in #516 will support this sort of packaging?

alexlarsson:
Its not generally possible to run something outside the sandbox like that. The whole point of the build is that we have all the build dependencies inside the runtime, and don't rely on host side tooling. You can however, disable the network sandboxing for a particular module, so that you can run something that downloads dependencies. (At least unless you specify --sandbox to flatpak-builder.)

Still, while this works, it doesn't really match how flatpak-builder normally works. Normally the json specifies the source up to sha256sums of every piece of sourcecode, and flatpak verifies that, so that the build is completely reproducible.

alexlarsson:
_From @lilianmoraru on February 15, 2017 11:47_

I've also hit this issue today.
I want to make my Rust server runnable only from Flatpak.

alexlarsson:
_From @daa84 on March 9, 2017 18:14_

[...]

Issue #494, opened 2022-10-15
More intelligent default Sdk extension version


hfiguiere:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a bug that matches the one I want to file, without success.

### flatpak-builder version

1.2.2

### Flatpak version

1.12.7

### How to reproduce

1. Use GNOME or KDE SDK
2. Add and extension to the manifest (rust-stable will work). Make sure it is NOT installed
3. Run flatpak-builder

### Expected Behavior

It offer to install the matching SDK extension, guessing the proper version, GNOME 43  -> version 22.08

### Actual Behavior

It tells me it cant find the SDK extension version 43

### Additional Information

The version is in the extension point of the SDK/Runtime.

#469 show we can't specify the SDK extension with a tuple (id//version) that could be an ok workaround

If the right SDK extension is already installed there is no problem.

TingPing:
This just sounds like a new feature. It doesn't really "guess" anything. It just defaults to the same version as the runtime.

hfiguiere:
From a user standpoint it's a bug. 

I don't know how feasible that is, but the determination of the version of the extension can be made by checking the extenion point for the `version` property.

Justinzobel:
A possible option (from my perspective) would be to have a version specified in the manifest:
```
    "sdk-extensions": [
        "org.freedesktop.Sdk.Extension.rust-stable//23.08"
    ],
```

hfiguiere:
> A possible option (from my perspective) would be to have a version specified in the manifest:
> 
> ```
>     "sdk-extensions": [
>         "org.freedesktop.Sdk.Extension.rust-stable//23.08"
>     ],
> ```

See #469

bbhtt:
Both this and https://github.com/flatpak/flatpak-builder/issues/469 would need to be fixed in Flatpak first. Flatpak builder just calls `flatpak build-init --sdk-extension=`. That a) does not allow specifying a branch and b) defaults to SDK branch.

Issue #469, opened 2022-05-09
[Bug]: References with version suffix don't work


ranfdev:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a bug that matches the one I want to file, without success.

### flatpak-builder version

1.2.2

### Flatpak version

1.12.7

### How to reproduce

Install an extension with a version suffix:
```
flatpak install --user org.freedesktop.Sdk.Extension.rust-stable//21.08
```

Create a manifest with that extension:
```
"sdk-extensions" : [
    "org.freedesktop.Sdk.Extension.rust-stable//21.08"
]
```

Run flatpak builder
```
flatpak-builder --user build manifest.json
```

### Expected Behavior

Given that the required extension is installed, flatpak-builder should find it without issues.

### Actual Behavior

```
error: Requested extension org.freedesktop.Sdk.Extension.rust-stable//21.08 not installed
```

PR #564, opened 2023-09-20
[WIP] Add Cargo buildsystem

chriskilding:
Adds a Flatpak buildsystem for Cargo.

Implements #15

---

Note: I am opening this PR at an early stage to gather initial feedback and requirements for what needs to be done. I'm not familiar with the internals of Flatpak, so will gratefully accept any help.


src/builder-module.c
```diff
@@ -1634,6 +1638,22 @@ builder_module_build_helper (BuilderModule  *self,
           return FALSE;
         }
     }
+  else if (cargo)
+    {
+      cargo_file = find_file_with_extension (source_subdir, "Cargo.toml");
```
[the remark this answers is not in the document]
chriskilding: Hmm, Cargo.lock only contains dependency definitions (and nothing else), so if the build process needed to read anything else like metadata (name / description / version ...) then it would need Cargo.toml to be present too

chriskilding:
I'm also a bit stuck on how to add the new test cases to the script / entry point that runs the test suite - could someone advise where this is done, so that they can be run in the build?

src/builder-module.c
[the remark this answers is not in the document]
nanonyme: Never mind, I mistook this as being related to cargo sources whereas this is actually about cargo build.

TingPing:
The entry point is https://github.com/flatpak/flatpak-builder/blob/main/tests/test-builder.sh

They are just simple shell scripts.

TingPing:
So we need to make this have value beyond what `buidlsystem: simple` just calling `cargo` does. That means that we have something that actually expands Cargo sources or similar.

chriskilding:
@TingPing agreed - that was always the intention. I just wanted to get the test harness set up around it first to TDD my changes.

Would you still recommend `tomlc99`, or are there any better (perhaps Cargo-specific) parser options?

TingPing:
I have no opinion at all of TOML libraries.

On Fedora these are pacakged:

- [libtoml](https://github.com/ajwans/libtoml) (C)
- [tomml11](https://github.com/ToruNiina/toml11) (C++ Header)
- [tomlplusplus](https://github.com/marzer/tomlplusplus) (C++)
- ~~[cpptoml](https://github.com/skystrife/cpptoml) (C++) (DEAD)~~

TingPing:
Not packaged but maybe interesting:

- [tomlc99](https://github.com/cktan/tomlc99) (C)
- ~~[tomlc](https://github.com/mzgoddard/tomlc) (C) (DEAD)~~

TingPing:
I think I'd lean towards tomlc99 as a Meson subproject.

chriskilding:
Another thought about parsing the Cargo / Cargo.lock files....

If we take the Toml parser route with `tomlc99` or similar, effectively we would have to reimplement all of Cargo's parse+fetch logic in C. Furthermore the Cargo project occasionally changes the Cargo.lock file format (as indicated by its version number), so each time this changes we'd have to update the `flatpak-builder` reimplementation. This will likely be non-trivial.

I am therefore thinking that either of the following strategies would be better than reimplementing the Cargo dependency fetch logic ourselves:

1. Use the `cargo` crate and some kind of Rust-C FFI bindings to call the fetch Rust function from C directly. (This presumes that this function exists and is publicly available within Cargo's internal library. I imagine the FFI bindings would make this option more difficult to implement.)
2. Do a shell call within `flatpak-builder` to effectively exec `cargo fetch`. (I think this will be simpler to implement, and also make it easier to accommodate newer versions of Cargo and the Cargo.lock file format. We would however have to take care to show any errors from the delegated `cargo fetch` call to the user.)

Thoughts?

src/builder-module.c
```diff
@@ -1585,6 +1587,8 @@ builder_module_build_helper (BuilderModule  *self,
     simple = TRUE;
   else if (!strcmp (self->buildsystem, "qmake"))
     qmake = TRUE;
+  else if (!strcmp (self->buildsystem, "cargo"))
+    cargo = TRUE;
```
hfiguiere, line 1591: this will need to be documented. See `data/flatpak-manifest.schema.json` and `doc/flatpak-manifest.xml`

src/builder-module.c
```diff
@@ -1741,6 +1761,10 @@ builder_module_build_helper (BuilderModule  *self,
               configure_cmd = "meson";
               configure_final_arg = g_strdup ("..");
             }
+          else if (cargo)
+            {
+              // nothing to do - Cargo does not have a configure step
```
hfiguiere, line 1766: so how do you select features?
or building a specific binary.

hfiguiere:
Also one key thing here is that `cargo` isn't part of the base sdk. So the failure mode might not be ideal.

src/builder-module.c
chriskilding: Good point. Just to be more specific about this, do you mean switches like:

- Profile: `cargo build --release / --debug`
- Target: `cargo build --target <architecture>`
- Cargo features: `cargo build --features <feature>,<feature>,...`

chriskilding:
@hfiguiere yes, also a good point. I suppose this prompts the question of how `flatpak-builder` should provide buildsystems for **any** language/toolchain that's not in the base SDK - not just Rust/Cargo.

alufers:
> "The buildsystem depends on the following sdk-extensions [ foo, bar, baz ], but these were not present. Please declare them in your Flatpak manifest."
> 
> Meanwhile a more automated option would be for `flatpak-builder` to just go ahead and resolve all the buildsystem's sdk-extension dependencies by itself. This would, however, mean that those sdk-extensions are not declared in the Flatpak manifest. There might be unintended consequences to that, but I don't have a view yet on what those could be.

The problem I see with this is that you can choose multiple different rust SDK extensions (stable, nightly etc.):

https://github.com/flathub/org.freedesktop.Sdk.Extension.rust-stable
https://github.com/flathub/org.freedesktop.Sdk.Extension.rust-nightly

To add insult  to injury, those SDK extensions put cargo in different paths, and they have to be appended to PATH manually in the manifest.

yonas:
How do I work around this until this feature is officially supported?

yonas:
@hfiguiere What's 1 + 1?

bbhtt:
Fwiw if there is an interest to work on this I wouldn't tie / get bogged down on the question that rust / cargo is missing from the default reference SDK.

That isn't very different to how things work on an distro or on CI (esp. containers). If people need something they have to install it one way or the other.

How they would do it is entirely upto them - they can use an SDK that comes with rust, put a rust module inside their manifest with cleanup set to `*`, use the extension or use rustup with network access in manifest and so on.

You also can't expect the SDK to have every exotic build tooling like say bazel available everytime we want to add a dedicated buildsystem module for it and progress can't get stuck on that.

Part of the UX/DX here will become easier if https://github.com/flatpak/flatpak-builder/issues/494 or #469 is solved.

There is also no need to make it the buildsystem API resolve dependencies.

The important thing to me here is that I can't find a suitable, actively maintained toml parser to depend upon. tomlc99 was a good candidate but seems to have been obsoleted in the meantime and replaced by something that is also not being actively maintained.

Also reimplementing part of Cargo here is a bad idea in general.

So I think it would be easier to shell to cargo from inside the build sandbox. That's also how meson cmake etc. are used.

bbhtt:
Also I agree with Patrick that a dedicated buildsystem needs to provide value beyond using simple otherwise there isn't a point to make it.

I don't see this PR at this moment doing that.

Providing value for me would be: if build sandbox has network acess buildsystem: cargo should locate appropriate lockfile or if there isn't one create it from Cargo.toml then fetch dependencies over network and run install steps. Install steps would need profile and feature selections etc.

If offline: go directly to the install step with `--frozen`.

Also job control with the `--jobs` argument and tests etc.

PR #566, opened 2023-10-21
Added more testing

hfiguiere:
- test the output of both yaml and json (one was discarded).
- test rename desktop and rename icon.


hfiguiere:
This is a preliminary to the PR I will submit to #1.

PR #567, opened 2023-10-22
Issue #1 - Rename mime files and icons

### LINKED ISSUE #1: Provide way to rename mime .xml files
_From @jurf on August 13, 2017 14:17_

So we can stop doing things like:

```
"post-install": [
    "mv /app/share/mime/packages/scribus.xml /app/share/mime/packages/net.scribus.Scribus.xml"
]
```

_Copied from original issue: flatpak/flatpak#958_

hfiguiere:
~~This currently sits on top of #566~~

Close #1


hfiguiere:
My concern is how to handle the compatibility. 

Currently you get these warnings:
```
** (flatpak-builder:1488560): WARNING **: 22:02:40.385: Unknown property rename-mime-file for type BuilderManifest

** (flatpak-builder:1488560): WARNING **: 22:02:40.385: Unknown property rename-mime-icons for type BuilderManifest
```

But that mean that the rename won't be done. Lots of packages do this manually.

hfiguiere:
Apparently the XML writing code in flaptak-builder is buggy.

hfiguiere:
Something like
```xml
            <match type="string" value="&lt;?xml" offset="0">
```
is rewritten as 
```xml
            <match type="string" value="<?xml" offset="0">
```

TingPing:
If its data you are adding call `g_markup_escape_text()` on it.

hfiguiere:
something like that.

This was from real-life testing.

TingPing:
> My concern is how to handle the compatibility.

I don't think there is anything we can really do about that, other than have Flathub update quickly.

TingPing:
Hmm. I can't reproduce with what you've pushed here, though I can certainly understand `attribute_values` being incorrect at some point.

It looks like `node->attribute_values` are always unescaped markup, and they get escaped at write by `g_markup_printf_escaped()`.

hfiguiere:
> Hmm. I can't reproduce with what you've pushed here

That's because I already pushed the fix. I even added it in the test case.

TingPing:
Oh excellent :)

The XML parsing does incorrectly handle values though:

```<match type="string" value="<?xml" offset="0">``` gets let through when it should error.

hfiguiere:
This push https://github.com/flatpak/flatpak-builder/compare/d2ac83a5f0cc7d80062f8da32ee316a1ff8fa19c..132c1816123f19ceca02a00c41a92ac62a1ea933

TingPing:
Err, nevermind. It does get escaped, but it is technically invalid input.

TingPing:
@hfiguiere What do you think of this: https://github.com/flatpak/flatpak-builder/commit/1709fb64b096307359b6e9185cf3c0904a44d111

hfiguiere:
> @hfiguiere What do you think of this: [0d2e76a](https://github.com/flatpak/flatpak-builder/commit/0d2e76a558bd9e09173d2e009ee9444126670ce1)

Looks ok to me. But my testing was done with files that validated, and the error was detectd by the mime-update, which the input supposedly passed.

It's not a dependency for this.

TingPing:
The change actually needs to happen in `GMarkupParser` I think, since it unescapes before it gets to this function. Anyway being overly relaxed on input probably isn't a big deal.

TingPing:
FWIW: https://github.com/flatpak/flatpak-builder/issues/568

If I find time I might look into some day.

hfiguiere:
> The validation change actually needs to happen in `GMarkupParser` I think, since it unescapes before it gets to this function. Anyway being overly relaxed on input probably isn't a big deal.

Usually it's relaxed on input and strict on output.

As discussed, in the longer term it might be wise to use libxml2 to handle XML, but that's not something blocking right now.

There error was triggered by `update-mime-database` that flatpak runs after installing.

JakobDev:
> My concern is how to handle the compatibility.

That's no problem. We already have flatpak-builder packaged as Flatpak, so everyone can use the latest version

PR #574, opened 2023-12-12
Print a soft deprecation warning when app-id is used

bbhtt:
See https://github.com/flatpak/flatpak-builder/commit/261575fb3912a7b0229b2427871ded9977340673


TingPing:
I don't want to be super verbose about it as it doesn't *really* matter.

Maybe in 1.5.x.

bbhtt:
Yea sure.

bbhtt:
I'm closing this, I don't think the warning is very useful since this is never going to be removed.

Adding it to the docs was enough.

PR #575, opened 2023-12-18
Add support for external dir hasher

abergmeier:
Allows providing hash-cmd.
Protocol is that cmd will be executed and output will be used to calculate
dir checksum.
If program fails - it falls back to rebuilding.

### DIFF
--- a/src/builder-source-dir.c
+++ b/src/builder-source-dir.c
@@ -36,6 +36,7 @@ struct BuilderSourceDir
 {
   BuilderSource parent;
 
+  char        **hash_cmd;
   char         *path;
   char        **skip;
 };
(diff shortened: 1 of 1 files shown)


TingPing:
Can you give some real world examples of hash commands for an entire directory?

Also it is *very* common that a project uses a flatpak build dir inside of the same directory, which is impossible to say nothing changes, so maybe there should be a check for that.

abergmeier:
The general question though is whether it would not be better to have a hash for every build-command...

TingPing:
I don't think it should be per build-command. It is a departure from how f-b operates.

bbhtt:
I will close this. There isn't much significant code here and it has bitrotten / stalled.

My initial instinct is also that I am not sold on the design of executing an external command for hashing and caching. I think if we do decide do it and some suitable standard method exists, the implementation should live inside flatpak-builder.

I also looked at a bunch of these tools that claim to support dir hashing and a. they are mostly not very actively maintained now b. they use different standards / methods and I am pretty sure it will lead to differences.

Issue #573, opened 2023-12-10
screenshots folder is no longer created in the builddir


tytan652:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a bug that matches the one I want to file, without success.

### flatpak-builder version

1.3.5

### Flatpak version

1.15.4

### How to reproduce

Use the [flatpak/flatpak-github-actions](https://github.com/flatpak/flatpak-github-actions) with a `mirror-screenshots-url` set.

### Expected Behavior

The build pass.

### Actual Behavior

The build fails:

```
Committing screenshots...
Uploading artifact...
Starting artifact upload
For more detailed logs during the artifact upload process, enable step-debugging: https://docs.github.com/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging#enabling-step-debug-logging
Artifact name is valid!
/usr/bin/ostree commit --repo=repo --canonical-permissions --branch=screenshots/x86_64 flatpak_app/screenshots
error: opendir(flatpak_app/screenshots): No such file or directory
/__w/_actions/flatpak/flatpak-github-actions/0ab9dd6a6afa6fe7e292db0325171660bf5b6fdf/flatpak-builder/dist/index.js:7912
                error = new Error(`The process '${this.toolPath}' failed with exit code ${this.processExitCode}`);
                        ^

Error: The process '/usr/bin/ostree' failed with exit code 1
    at ExecState._setResult (/__w/_actions/flatpak/flatpak-github-actions/0ab9dd6a6afa6fe7e292db0325171660bf5b6fdf/flatpak-builder/dist/index.js:7912:25)
    at ExecState.CheckComplete (/__w/_actions/flatpak/flatpak-github-actions/0ab9dd6a6afa6fe7e292db0325171660bf5b6fdf/flatpak-builder/dist/index.js:7895:18)
    at ChildProcess.<anonymous> (/__w/_actions/flatpak/flatpak-github-actions/0ab9dd6a6afa6fe7e292db0325171660bf5b6fdf/flatpak-builder/dist/index.js:7789:27)
    at ChildProcess.emit (node:events:514:28)
    at maybeClose (node:internal/child_process:1105:16)
    at ChildProcess._handle.onexit (node:internal/child_process:305:5)
```

[...]

PR #576, opened 2023-12-28
builder-manifest: run appstream compose with --no-partial-urls

hfiguiere:
Only when mirroring. This is to be compatible with existing flatpak deployment and appstream-glib

See https://github.com/ximion/appstream/issues/505

Should fix https://github.com/flatpak/flatpak-builder/issues/573

This can be tested on flathub with https://github.com/flathub/org.flatpak.Builder/pull/196


bbhtt:
Seems to work. This requires >=0.16.4 in [meson.build](https://github.com/flatpak/flatpak-builder/blob/7d621d6eec7fe523c68d46f3853cdc86561b5637/meson.build#L44)

barthalion:
This fixes a problem we would experience in the Flathub backend, and `--no-partial-urls` was added per our request.

TingPing:
> This requires at least appstreamcli>=0.16.3 in [meson.build](https://github.com/flatpak/flatpak-builder/blob/7d621d6eec7fe523c68d46f3853cdc86561b5637/meson.build#L44).

I think this is all that's needed to land this.

hfiguiere:
> ../meson.build:44:15: ERROR: Invalid version of program, need 'appstreamcli' ['>= 0.16.3'] found '0.16.1'.

I guess building on fedora 38 is a no go.

hfiguiere:
Also it's a runtime dependency while this is a meson (build-time) check.

hfiguiere:
done, but I don't think it will build in CI

hfiguiere:
On CI:
```
Program appstreamcli found: NO found 0.15.2 but need: '>= 0.16.3' (/usr/bin/appstreamcli)
```

TingPing:
Ubuntu 23.10 has 0.16.2... how annoying.

hfiguiere:
I'll reset the checks in meson and check at runtime in some way. Using CLI as an API...

hfiguiere:
The approach here would be to degrade if the version of appstream is too low. But the check is a bit too hackish.

Checking in meson.build is kind of irrelevant since it's build time.

PR #579, opened 2024-01-20
manifest: Properly change the app id in the appstream file

### LINKED ISSUE #90: Renaming an applications id should add provides/id to appstream file
If you rename foo.desktop and foo.appdata.xml to org.bar.Foo then it should label such in the appdata file so software can better show this.

```xml
<provides>
  <id>foo.desktop</id>
</provides>
```

hfiguiere:
It include more testing

Close #90

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -2787,22 +2787,29 @@ builder_manifest_cleanup (BuilderManifest *self,
             }
         }
 
-      if (self->rename_desktop_file != NULL)
+      if (self->rename_desktop_file != NULL || self->rename_appdata_file != NULL)
         {
-          g_autoptr(GFile) applications_dir = g_file_resolve_relative_path (app_root, "share/applications");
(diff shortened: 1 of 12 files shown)


src/builder-manifest.c
```diff
@@ -2827,13 +2832,27 @@ builder_manifest_cleanup (BuilderManifest *self,
...
                     }
                 }
 
+              n_provides = flatpak_xml_find (n_root, "provides", NULL);
+              if (!n_provides)
+                {
+                  n_provides = flatpak_xml_new ("provides");
+                  flatpak_xml_add (n_root, n_provides);
+                }
+              if (n_provides)
```
TingPing, line 2848: Will this always be true?

hfiguiere: likely. But I try to avoid null derefences.

hfiguiere: but then the same checks aren't done elsewhere.

TingPing: `flatpak_xml_new` can't fail and always returns something.

PR #583, opened 2024-02-16
manifest: Don't emit empty id in `provides` when renaming

hfiguiere:
Followup from https://github.com/flatpak/flatpak-builder/pull/579


TingPing:
Isn't a file without an id just invalid?

hfiguiere:
this is just "we don't need to rename", so we shouldn't add a `provides`.

See https://github.com/flatpak/flatpak-builder/pull/583/files#diff-81460d46a2933ba7be79c8a9e9b7df7c28c7b6d867a80ae5436bbafd9108dbdaR2834

PR #584, opened 2024-02-20
Fix installed-tests failure

smcv:
* tests: Install test-rename-appdata.json for "as-installed" testing

---

To reproduce: install https://gitlab.gnome.org/GNOME/gnome-desktop-testing, build with `-Dinstalled_tests=true`, install, run `gnome-desktop-testing-runner flatpak-builder`. We use this in the Debian packaging, to catch potential bugs where upgrading a dependency might break flatpak-builder.

Not tested yet, therefore not necessarily a complete fix. I'll mark as non-draft when ready.


TingPing:
Would you like a 1.4.2 with this fix?

smcv:
It might make sense to have a new release for this and d9f5a5e. For now I applied both as patches in Debian.

PR #587, opened 2024-02-24
Default libdir for app is /app/lib

nanonyme:
flatpak-builder promises that libdir is /app/lib for apps but this is currently not really the case. Instead, flatpak-builder relies on patching of libdir on Meson and cmake and prefix being set to /app. Since it's not really up to runtime to select what libdir for apps is (this is defined in flatpak instead), instead add a default here.


nanonyme:
I will close this for further work.

PR #588, opened 2024-02-24
Always calculate libdir

nanonyme:
flatpak-builder relies on patched cmake and meson from
freedesktop-sdk to change libdir from lib64 to lib on 64bit
distributions. This dependency is *weird* since it's actually
flatpak that sets up the dynamic linker paths. Instead, put
responsibility for setting up correct libdir to flatpak-builder
always.


nanonyme:
Not sure whether there's a risk that always passing libdir breaks something.

TingPing:
I think this is fine.

I tested cmake, autotools, and meson. If you pass libdir twice the second one always wins, so manifests with custom values will still be respected.

nanonyme:
It's not actually set twice. New logic is:
1. If module sets libdir, use that
2. If building runtime, read from runtime configuration file and use that
3. If libdir is completely unset, calculate "lib" directory under prefix.

The step 3 is new, steps 1 and 2 were there already.

TingPing:
I don't mean *we* set it twice. Just an old manifest may have set it manually itself.

PR #592, opened 2024-03-02
Move locale migration after cleanup

### LINKED ISSUE #154: Strange behaviour when cleaning up "/share/locale"
Take the following example _org.test.test.json_ file
```json
{
    "app-id": "org.test.test",
    "runtime": "org.gnome.Platform",
    "runtime-version": "3.28",
    "sdk": "org.gnome.Sdk",
    "modules": [
        {
            "name": "testA",
            "cleanup": [
                "/share/locale"
            ],
            "buildsystem": "simple",
            "build-commands": [
                "mkdir -p /app/share/locale/es",
                "touch /app/share/locale/es/testA.mo"
            ]
        },
        {
            "name": "testB",
            "buildsystem": "simple",
            "build-commands": [
                "mkdir -p /app/share/locale/es",
                "touch /app/share/locale/es/testB.mo"
            ]
        }
    ]
}
```
Running `$ flatpak-builder dir --repo=repo org.test.test.json` creates the directories `dir/files/share/runtime`. But there is no `dir/files/share/locale/es/testB.mo`.

After removing the cleanup part `flatpak-builder` creates `dir/files/share/locale/es/test{A|B}.mo` as expected.

It only happens for `/share/locale`, maybe because it is treated special.

chrisawi:
Otherwise, module-level cleanup can remove locale symlinks that point to locale data installed by other modules.

Closes #154

I've tested this building an app, and for that at least, it seems to work. I haven't tried building a runtime though. There are two other calls to `builder_migrate_locale_dirs()` that I haven't really investigated. One happens in init for all runtimes, and the other applies only to platforms.

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -2761,6 +2761,22 @@ builder_manifest_cleanup (BuilderManifest *self,
             }
         }
 
+      if (builder_context_get_separate_locales (context))
+      {
+        g_autoptr(GFile) root_dir = NULL;
+
+        if (builder_context_get_build_runtime (context))
+          root_dir = g_file_get_child (app_dir, "usr");
+        else
+          root_dir = g_file_get_child (app_dir, "files");
+
+        if (!builder_migrate_locale_dirs (root_dir, error))
+          {
+            g_prefix_error (error, "Can't migrate locale dirs: ");
+            return FALSE;
+          }
+      }
+
       app_root = g_file_get_child (app_dir, "files");
 
       appdata_source = builder_manifest_find_appdata_file (self, app_root);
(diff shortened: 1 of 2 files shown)


TingPing:
If you could, add some tests to https://github.com/flatpak/flatpak-builder/blob/main/tests/test-builder.sh

src/builder-manifest.c
```diff
@@ -2783,6 +2783,22 @@ builder_manifest_cleanup (BuilderManifest *self,
             }
         }
 
+      if (builder_context_get_separate_locales (context))
+      {
```
swick: indentation is off here

PR #601, opened 2024-05-19
build-commands: add: The build process fails when an individual entry…

nandedamana:
… in the array fails.

Just making it clear so that one won't be hesitant to break down a long '&&'-ed one-liner.

### DIFF
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -576,6 +576,8 @@
                     <listitem><para>An array of commands to run during build (between make and make install if those are used).
                     This is primarily useful when using the "simple" buildsystem.
                     Each command is run in <literal>/bin/sh -c</literal>, so it can use standard POSIX shell syntax such as piping output.
+                    If any individual entry in the array fails, then the whole build process will fail,
+                    similar to commands in a <citerefentry><refentrytitle>make</refentrytitle><manvolnum>1</manvolnum></citerefentry> recipe.
                     </para></listitem>
                 </varlistentry>
                 <varlistentry>


smcv (changes requested):
Thanks, it would be good to document this behaviour.

The first line of the commit message is rather long (and Github is giving you a hint that this might be a problem by ellipsizing it). Maybe something like this?

```
flatpak-manifest(5): If one command fails, the whole build fails

Documenting this makes it more obvious that it's valid to write

    - foo
    - bar
    - baz

and unnecessary to combine those commands into
`foo && bar && baz` to get sensible error behaviour.
```

smcv:
> Just making it clear so that one won't be hesitant to break down a long '&&'-ed one-liner.

This is good reasoning that ought to be captured in the commit message for the benefit of future maintainers. (For example see https://cbea.ms/git-commit/)

smcv:
Please squash your two commits into one, ideally with an explanatory commit message like the one I suggested in https://github.com/flatpak/flatpak-builder/pull/601#pullrequestreview-2066126671

smcv:
Yeah, it's clearly a bug that flatpak-builder doesn't run `false` as a command in that situation, but also doesn't error out with some semi-useful message like "expected string, but found boolean". Please report that as an issue if nobody has done so already - and I'm sure a PR to solve it would be appreciated, if you have some time.

`false` is not even the worst YAML trap: if you're discussing ISO country codes, it's easy to get tripped up by Norway being `no`, which (if unquoted) also gets interpreted as boolean FALSE ("the Norway problem").

nandedamana:
> Please squash your two commits into one, ideally with an explanatory commit message like the one I suggested in [#601 (review)](https://github.com/flatpak/flatpak-builder/pull/601#pullrequestreview-2066126671)

Your commit message looked fine, so I just reused it.

nandedamana:
@smcv Anything else to be done from my part?

PR #608, opened 2024-07-01
Split autotools and meson

nanonyme:
autotools is not really autotools but covers arbitrary legacy build systems using make. It's not safe to touch arguments there.


nanonyme:
I guess the fact tests needed changing in previous PR should have rung alarm bells.

nanonyme:
@TingPing do you have time to look at this?

TingPing:
I don't have a super strong opinion.

I think Perl is a rare example of a configure script that doesn't support this and the `simple` build type exists for anything that falls out of scope.

bbhtt:
The other issue with https://github.com/flatpak/flatpak-builder/pull/588 is `CMAKE_INSTALL_LIBDIR` is not an absolute path it is being set to `/app/lib` when it should've been `lib`. `CMAKE_INSTALL_FULL_LIBDIR` is supposed to be absolute.

This is now causing https://github.com/flathub/io.github.martinrotter.rssguard/issues/73#issuecomment-2208189029

TingPing:
Can you submit a PR for that?

TingPing:
`gperftools` is another example of breaking configure:

`configure: error: expected an absolute directory name for --libdir: lib`

PR #609, opened 2024-07-02
tests/testpython.py: Covert to Python 3

travier:
See: https://bugzilla.redhat.com/show_bug.cgi?id=2249819
See: https://fedoraproject.org/wiki/Changes/RetirePython2.7

### DIFF
--- a/tests/testpython.py
+++ b/tests/testpython.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/python
 
 import importme
-print importme.foo
+print(importme.foo)


travier:
Untested.

tests/testpython.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/python
 
 import importme
 print importme.foo
```
hroncok, line 4: This is Python 2 syntax.

hroncok:
I've read https://github.com/flatpak/flatpak-builder/commit/9c57a5751ee5c0403700c9da0d35ba56b08c1995 and porting this file to Python 3 makes very little sense to me.

hroncok:
To clarify: The purpose of the test seems to be a script with python2 shebang.

If you change the shebang (and port the script to python 3), the purpose of the test is defeated.

The test is skipped when python2 is not available. To fix https://bugzilla.redhat.com/show_bug.cgi?id=2249819 IMHO the Fedora package should filter the automatic dependency out instead.

debarshiray:
> To clarify: The purpose of the test seems to be a script with python2 shebang.
> 
> If you change the shebang (and port the script to python 3), the purpose of the test is defeated.

Thanks for digging into the code.  I never got around to digging into what commit 9c57a5751ee5c0403700c9da0d35ba56b08c1995 and commit 006d9a19270416f7c57b7ba4220ef8ba306e8b94 were actually trying to accomplish, which was made worse by my almost non-existent knowledge of Python these days.  So, thanks for doing that.

> The test is skipped when python2 is not available. To fix https://bugzilla.redhat.com/show_bug.cgi?id=2249819 IMHO the Fedora package should filter the automatic dependency out instead.

Okay!

debarshiray (commented):
If this commit is *really* needed, then there's a minor typo in the commit message.  It should be *convert*, not *covert*.  :)

TingPing:
I'll close this since it is a python2 specific test. Someday it can just be removed entirely but it costs nothing to leave it there, and `python2` can and will still be used in flatpak for a least a bit.

travier:
👍🏻 Thanks for digging!

PR #610, opened 2024-07-05
builder-module: Fix cmake libdir value

bbhtt:
Fixes 8c036e00630e35423c03388aacc06cd00dda74ea

It is being set to `$prefix/lib` but CMAKE_INSTALL_LIBDIR is not an absolute path by convention [1]:

> It should be a path relative to the installation prefix

When it is set to `/app/lib`, it is causing issues in applications [2]

[1]: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables

[2]: https://github.com/flathub/io.github.martinrotter.rssguard/issues/73#issuecomment-2208189029

### DIFF
--- a/src/builder-module.c
+++ b/src/builder-module.c
@@ -1788,7 +1788,7 @@ builder_module_build_helper (BuilderModule  *self,
           if (libdir)
             g_ptr_array_add (configure_args_arr, g_strdup_printf ("-DCMAKE_INSTALL_LIBDIR:PATH='%s'", libdir));
           else
-            g_ptr_array_add (configure_args_arr, g_strdup_printf ("-DCMAKE_INSTALL_LIBDIR:PATH='%s/lib'", prefix));
+            g_ptr_array_add (configure_args_arr, g_strdup_printf ("-DCMAKE_INSTALL_LIBDIR:PATH='lib'"));
           g_ptr_array_add (configure_args_arr, g_strdup ("-G"));
           g_ptr_array_add (configure_args_arr, g_strdup_printf ("%s", cmake_generator));
         }


src/builder-module.c
```diff
@@ -1788,7 +1788,7 @@ builder_module_build_helper (BuilderModule  *self,
           if (libdir)
             g_ptr_array_add (configure_args_arr, g_strdup_printf ("-DCMAKE_INSTALL_LIBDIR:PATH='%s'", libdir));
```
TingPing, line 1789: This `libdir` can also be an absolute path.

bbhtt: That's user input though, if they set wrong values they will get weird behaviour back.

TingPing: The docs say the default is `/app/lib`. It was clearly intended to support absolute paths and it should be fine to accommodate both.

bbhtt: So what to do? You can already set that to an absolute path if required for some reason in the manifest.

TingPing:
```c
          if (libdir && g_path_is_absolute (libdir))
            g_ptr_array_add (configure_args_arr, g_strdup_printf ("-DCMAKE_INSTALL_FULL_LIBDIR:PATH='%s'", libdir));
          else if (libdir)
            g_ptr_array_add (configure_args_arr, g_strdup_printf ("-DCMAKE_INSTALL_LIBDIR:PATH='%s'", libdir));
```

bbhtt: Note that setting CMAKE_INSTALL_FULL_LIBDIR, will have no effect on the build if it is not being used as a variable in cmakelists.txt/in other places. I've seen very few software use the FULL paths in pkgconfig or cmakelists.

bbhtt: CMAKE_INSTALL_LIBDIR is more universally used, there is also stuff that uses it wrongly as an absolute path too.

TingPing: Hmm that is unfortunate.

Anyway this PR is fine, that was a pre-existing issue.

PR #614, opened 2024-07-31
Install Debug and Locale extension by default

ColinKinloch:
Alternative to #520 based on feedback.

I'm not sure how to check whether it should attempt to install Debug and Locale extensions, if someone with more experience could chip in I'd be grateful.


bbhtt:
Do you still plan to work on this? It's exporting the Debug and Locale a bit above, you can just check from that?

PR #625, opened 2024-11-23
Don't ignore `--build-shell` option when configuration fails

### LINKED ISSUE #479: [Bug]: --build-shell parameter stops after configuration not download/unpackaging

nbenitez:
`--build-shell` is supposed to help in debbuging problems for modules, that may be a failure to build, to configure, etc.

So it doesn't make much sense that if the configuration step fails we ignore the `--build-shell` option, which the developer is probably using it to debug that same configure problem.

Fixes #479

### DIFF
--- a/src/builder-module.c
+++ b/src/builder-module.c
@@ -1590,7 +1590,7 @@ builder_module_build_helper (BuilderModule  *self,
     {
       g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "module %s: Invalid buildsystem: \"%s\"",
                    self->name, self->buildsystem);
-      return FALSE;
+      return run_shell ? shell (app_dir, self->name, context, source_dir, build_dir_relative, build_args, env, error) : FALSE;
     }
 
   if (simple)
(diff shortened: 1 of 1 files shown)


nbenitez:
PR #508 tried to fix this bug, but @TingPing suggested there in [a comment](https://github.com/flatpak/flatpak-builder/pull/508#issuecomment-1369317784) to better use another approach which I implemented in this PR.

src/builder-module.c
```diff
@@ -1590,7 +1590,7 @@ builder_module_build_helper (BuilderModule  *self,
     {
       g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "module %s: Invalid buildsystem: \"%s\"",
                    self->name, self->buildsystem);
-      return FALSE;
+      return run_shell ? shell (app_dir, self->name, context, source_dir, build_dir_relative, build_args, env, error) : FALSE;
```
swick, line 1593: This is passing in a non-empty error, so that won't work.

PR #636, opened 2025-03-19
docs: Mention nested modules don't inherit `build-options`

salim-b:
I think it makes sense to explicitly mention this since it's kinda counterintuitive to what one might expect.

### DIFF
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -569,7 +569,7 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>build-options</option> (object)</term>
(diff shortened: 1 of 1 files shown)


TingPing:
Its harmless for this specific option so I don't mind, but its more of a general concept that module scopes are isolated. Only the global one is inherited.

salim-b:
> but its more of a general concept that module scopes are isolated

See #637 for a follow-up addition.

Another question: Would it be possible somehow to display a table of contents for the "Command Reference" pages, i.e. have a floating "On this page" ToC on the right side of the rendered page? Without it, it's really hard to keep an overview...

PR #637, opened 2025-03-20
docs: Be explicit about module scope isolation

salim-b:
Follow-up to #636

### DIFF
--- a/doc/flatpak-manifest.xml
+++ b/doc/flatpak-manifest.xml
@@ -483,7 +483,7 @@
                 the build options and a list of sources to download and extract before building.
             </para>
             <para>
-                Modules can be nested, in order to turn related modules on and off with a single key.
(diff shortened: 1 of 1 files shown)


doc/flatpak-manifest.xml
```diff
@@ -483,7 +483,7 @@
                 the build options and a list of sources to download and extract before building.
             </para>
             <para>
-                Modules can be nested, in order to turn related modules on and off with a single key.
+                Modules can be nested, in order to turn related modules on and off with a single key. Module scopes are always isolated, so nested modules do <emphasis>not</emphasis> inherit attributes from parent modules.
```
bbhtt: this should probably be in a new line and i'd drop the emphasis.

salim-b: I've dropped the emphasis. But I prefer to keep it in the same paragraph (it belongs together since it refers to the nesting).

bbhtt: No i was saying it for the line length, but there are longer lines already so it is ok i guess.

swick:
Should be good to merge, no?

bbhtt:
Yea.

PR #643, opened 2025-05-20
builder-module: Change the default wrap-mode policy of meson

alatiera:
By default network access inside the build sandbox is disabled, however meson defaults into trying to initialize fallbacks and attempt to download them, returning a potentially misleading error message.

Setting wrap-mode to nodownload will instruct meson to not try to use the network to fetch any resources and only the pre-existing checkouts will be taken into account. This matches both the default network permission and the design of flatpak-builder for bundling and staging all the sources ahead of time.

Side effect is that it makes it slightly harder for projects to use the network access at build time, is they also have to edit their meson setup arguments to change the wrap-mode.


src/builder-module.c
```diff
@@ -1780,6 +1780,22 @@ builder_module_build_helper (BuilderModule  *self,
...
+          * download them, returning a potentially misleading error message.
+          *
+          * Setting wrap-mode to nodownload will instruct meson to not try to use the
+          * network to fetch any resources and only the pre-existing checkouts will be
+          * taken into account. This matches both the default network permission
+          * and the design of flatpak-builder for bundling and staging all the sources
+          * ahead of time.
+          *
+          * Side effect is that it makes it slightly harder for projects to use the
+          * network access at build time, is they also have to edit their meson setup
```
alatiera, line 1795:
```suggestion
          * network access at build time, as they also have to edit their meson setup
```

bbhtt:
as is it will break ci uses cases of flatpak-builder where often the developer workflow is to build with network access. so probably this should not be set when the build sandbox has network access.

Issue #456, opened 2022-01-25
[Bug]: username:password from before @ in https URL not used


jangernert:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a bug that matches the one I want to file, without success.

### flatpak-builder version

1.2.0

### Flatpak version

1.12.3

### How to reproduce

1. Add a source of type `file` with an url that contains basic auth information to a `sources.json`
2. add the `sources.json` to the `sources` of the flatpak manifest
3. run flatpak-builder CLI

### Expected Behavior

downloading all the sources

### Actual Behavior

Failed to download sources: module: The requested URL returned error: 401

### Additional Information

The URL points to a gitlab nuget repository and contains basic auth information:

`https://user:password@gitlab.com/api/v4/projects/project_id/packages/nuget/download/package_name/package_version/file_name.nupkg`

Downloading the file via curl and firefox works fine.

smcv:
I suspect this means that flatpak-builder's downloading code doesn't support putting a username/password into the URL like this, so it is effectively using `https://gitlab.com/api/v4/projects/project_id/packages/nuget/download/package_name/package_version/file_name.nupkg`, which fails with error 401, meaning authentication is required.

jangernert:
That's what I suspected at first. But libsoup seems to support it. There are at least tests for it: https://github.com/GNOME/libsoup/blob/master/tests/auth-test.c#L1726

PR #646, opened 2025-06-21
Enable authenticated downloads via netrc file

Corristo:
Sometimes source archives are only accessible to logged-in users, for example as described in issue #456. However, authentication currently is not supported by flatpak-builder aside from hard-coding credentials in the manifest file.

Since flatpak-builder uses libcurl to download sources we can make use of cURL's support for netrc files to provide credentials out-of-band. While that does not directly solve #456 it offers an in my opinion superior alternative: Instead of hard-coding username and password in the manifest file the reporter of #456 can now create the file `~/.netrc` with the following content:

```
machine gitlab.com
  login <user>
  password <password>
```

This allows storing the manifest file in version control without exposing the credentials to all other developers.


bbhtt:
Seems to work. imo this should solve https://github.com/flatpak/flatpak-builder/issues/456, using username@pass in the manifest should not be allowed.

PR #649, opened 2025-08-03
builder-main: Commit mirrored media as screenshot ref to exported repo

bbhtt:
```
This is needed to store the application icons and the mirrored
screenshots (as done by appstream) to the ostree repo. This can be
useful in many ways, the primary being flat-manager extracts [1] these
screenshot refs and stores them in the server so that application
store frontends can use their own mirrored screenshot and icon URLs.

This has been for the longest time done at various downstream levels
such as in buildbot [2], flatpak-github-actions [3] and vorarbeiter [4]
(the successor to buildbot) but it's more user friendly to have
flatpak-builder do this extra work.

[1]: https://github.com/flatpak/flat-manager/commit/eff43004756c011b0a0d034a61cec84836b8e0a9
[2]: https://github.com/flathub-infra/buildbot/blob/9d8e2b5483b2e446389548446ebf9c3934380682/master/buildbot/flathub_master.py#L1081
[3]: https://github.com/flatpak/flatpak-github-actions/blob/6684584b07d86113b1acfc80b3e4667f16f617c3/flatpak-builder/index.js#L280-L286
[4]: https://github.com/flathub-infra/vorarbeiter/blob/7f2f04a562e73ba4f46044dae7c99f8e9e64a39d/justfile#L185-L189
```

PR #652, opened 2025-08-06
builder-manifest: Run appstream-compose with no-partial-urls argument

bbhtt:
Alternative to https://github.com/flatpak/flatpak-builder/pull/576

We can optionally pick up the meson/autotools bits from there, just to signal packagers that we ideally want >= 0.16.3


bbhtt:
For optional testability, I need the last commit of https://github.com/flatpak/flatpak-builder/pull/649/commits/7614f11db63bd09d9f247721e3cca3b140fc94d2 merged.

bbhtt:
> However, the generated output isn't something you generally should rely on

Yea we don't rely on that.

> That said, I can pretty much guarantee that the output of --version won't ever change.

Ok thanks.

> There is one caveat you need to be aware of: If, for whatever reason, you are using an appstreamcli built for one libappstream version with a libappstream that has a different version, passing that command will print both version numbers, which might confuse your parsing.

Yea I'm handling that by choosing the library version.

image

bbhtt:
ah i didn't knew about the --allow-custom flag.

bbhtt:
I tried the initial port to API here https://github.com/flatpak/flatpak-builder/pull/664 but looks like there the private asc-fonts header is not available, so it doesn't compile by default.

ximion:
> I tried the initial port to API here #664 but looks like there the private asc-fonts header is not available, so it doesn't compile by default.

Upgrade to AppStream 1.0.6 or higher and that issue will be gone. Nothing in the public API uses anything from that header anyway, so alternatively you can also just drop that reference or provide an empty header.

src/builder-manifest.c
```diff
@@ -3056,9 +3056,16 @@ builder_manifest_cleanup (BuilderManifest *self,
               g_autofree char *arg_media_dir =  g_strdup_printf ("--media-dir=%s",
                                                                  flatpak_file_get_path_cached (media_dir));
 
+              gboolean has_no_partial_urls = appstream_version_check (0, 16, 3);
+              if (!has_no_partial_urls)
+                {
+                  g_warning ("Found AppStream version < 0.16.3. Failed to add '--no-partial-urls' to compose");
+                }
```
swick:
```suggestion
              gboolean has_no_partial_urls = appstream_version_check (0, 16, 3);

              if (!has_no_partial_urls)
                g_warning ("Found AppStream version < 0.16.3. Failed to add '--no-partial-urls' to compose");
```

src/builder-utils.c
```diff
@@ -1836,3 +1836,50 @@ flatpak_version_check (int major,
...
+      as_micro == 0)
+    {
+      const char * argv[] = { "appstreamcli", "--version", NULL };
+      g_autoptr(GSubprocess) subp = NULL;
+      g_autofree char *out = NULL;
+
+      subp = g_subprocess_newv (argv, G_SUBPROCESS_FLAGS_STDOUT_PIPE, NULL);
+      g_subprocess_communicate_utf8 (subp, NULL, NULL, &out, NULL, NULL);
+
+      gchar **lines = g_strsplit (out, "\n", -1);
```
swick:
```suggestion
      g_auto(GStrv) lines = g_strsplit (out, "\n", -1);
```

But should also be declared at the top.

src/builder-utils.c
```diff
@@ -1836,3 +1836,50 @@ flatpak_version_check (int major,
...
+      for (int i = 0; lines[i] != NULL; i++)
```
swick:
```suggestion
      for (size_t i = 0; lines[i] != NULL; i++)
```
Nit picky, but indexing should really use size_t.

bbhtt:
I will get back to this a bit later. I realised this needs to be behind an arg `--compose-no-partial-urls` otherwise it breaks the behaviour in the stable release line. Then we can make a new 1.4.x release with this.

And drop it later for 1.5.0 or later when the appstream api is done.

src/builder-manifest.c
bbhtt: Ended up dropping this part.

src/builder-utils.c
bbhtt: applied thanks.

src/builder-utils.c
bbhtt: applied thanks.

bbhtt:
It should be ready for review now.

src/builder-main.c
```diff
@@ -144,6 +145,7 @@ static GOptionEntry entries[] = {
   { "assumeyes", 'y', 0, G_OPTION_ARG_NONE, &opt_yes, N_("Automatically answer yes for all questions"), NULL },
   { "no-shallow-clone", 0, 0, G_OPTION_ARG_NONE, &opt_no_shallow_clone, "Don't use shallow clones when mirroring git repos", NULL },
   { "override-source-date-epoch", 0, 0, G_OPTION_ARG_INT64, &opt_source_date_epoch, "Use this timestamp to perform the build, instead of the last modification time of the manifest.", NULL },
+  { "compose-url-policy", 0, 0, G_OPTION_ARG_STRING, &opt_as_url_policy, "Set the AppStream compose URL policy ('partial' or 'full'). Partial if unspecified.", "POLICY" },
```
swick:
```suggestion
  { "compose-url-policy", 0, 0, G_OPTION_ARG_STRING, &opt_as_url_policy, "Set the AppStream compose URL policy to either 'partial' (default) or 'full'", "POLICY" },
```

src/builder-main.c
```diff
@@ -608,6 +610,28 @@ main (int    argc,
...
+            }
+          builder_context_set_as_url_policy (build_context, BUILDER_AS_URL_POLICY_FULL);
+        }
+      else if (g_strcmp0 (opt_as_url_policy, "partial") == 0)
+        builder_context_set_as_url_policy (build_context, BUILDER_AS_URL_POLICY_PARTIAL);
+      else
+        {
+          g_printerr ("Invalid value for --compose-url-policy: %s\n", opt_as_url_policy);
+          return 1;
+        }
```
swick:
```suggestion
      BuilderAsUrlPolicy policy = BUILDER_AS_URL_POLICY_PARTIAL;

      if (g_strcmp0 (opt_as_url_policy, "full") == 0)
        policy = BUILDER_AS_URL_POLICY_FULL;
      else if (g_strcmp0 (opt_as_url_policy, "partial") == 0)
        policy = BUILDER_AS_URL_POLICY_PARTIAL;
      else if (opt_as_url_policy != NULL)
        {
          g_printerr ("Invalid value for --compose-url-policy: %s\n", opt_as_url_policy);
          return 1;
        }

      if (policy == BUILDER_AS_URL_POLICY_FULL && !appstream_version_check (0, 16, 3))
        {
          g_printerr ("AppStream version >= 0.16.3 required for 'full' compose URL policy\n");
          return 1;
        }

      builder_context_set_as_url_policy (build_context, policy);
```

src/builder-manifest.c
```diff
@@ -2423,6 +2423,7 @@ cmpstringp (const void *p1, const void *p2)
 
 static gboolean
 appstreamcli_compose (GError **error,
+                      BuilderAsUrlPolicy as_url_policy,
```
swick: Let's stick with the modern formatting.

src/builder-manifest.c
```diff
@@ -3067,6 +3071,8 @@ builder_manifest_cleanup (BuilderManifest *self,
           const char *opt_mirror_screenshots_url = builder_context_get_opt_mirror_screenshots_url (context);
           gboolean opt_export_only = builder_context_get_opt_export_only (context);
 
+          BuilderAsUrlPolicy as_url_policy = builder_context_get_as_url_policy (context);
+
```
swick, line 3074:
```suggestion
          BuilderAsUrlPolicy as_url_policy = builder_context_get_as_url_policy (context);

```

src/builder-context.h
```diff
@@ -29,6 +29,11 @@
 
 G_BEGIN_DECLS
 
+typedef enum {
+  BUILDER_AS_URL_POLICY_PARTIAL = 0,
+  BUILDER_AS_URL_POLICY_FULL
```
swick:
```suggestion
  BUILDER_AS_URL_POLICY_FULL,
```
Avoids having to change the line when a new value is added

src/builder-utils.h
```diff
@@ -216,6 +216,10 @@ gboolean flatpak_version_check (int major,
                                 int minor,
                                 int micro);
 
+gboolean appstream_version_check (int major,
```
swick: The name doesn't tell me what it checks. Does it check that it is at least the version specified?

src/builder-utils.c
```diff
@@ -1837,3 +1837,52 @@ flatpak_version_check (int major,
...
+            {
+              if (sscanf (lines[i], "AppStream version: %d.%d.%d", &as_major, &as_minor, &as_micro) == 3)
+                break;
+            }
+        }
+
+      if (as_major == 0 && as_minor == 0 && as_micro == 0)
+        g_warning ("Failed to find appstream version");
+
+      g_debug ("Found AppStream version %d.%d.%d", as_major, as_minor, as_micro);
```
swick:
```suggestion
      if (as_major == 0 && as_minor == 0 && as_micro == 0)
        g_warning ("Failed to find appstream version");
      else
        g_debug ("Found AppStream version %d.%d.%d", as_major, as_minor, as_micro);
```

tests/test-builder.sh
```diff
@@ -144,3 +146,15 @@ ostree checkout --repo=$REPO/repo_sc -U screenshots/$(flatpak --default-arch) ou
 find outdir_sc -path "*/icons/64x64/org.test.Hello.png" -type f | grep -q .
 
 echo "ok screenshot ref commit"
+
+# test compose partial url
+${FLATPAK_BUILDER} --force-clean builddir_sc \
+    --mirror-screenshots-url=https://example.org/media \
+    --state-dir .fp-compose-url-policy \
+    --compose-url-policy=partial \
```
swick, line 154: Maybe also add a negative test and one with policy=full?

src/builder-utils.h
bbhtt: Yea, if the run time appstream version >= (maj, min, micro) that we asked for.

tests/test-builder.sh
bbhtt: it requires appstream 0.16.3 at least which is not in ci and also needs guarding via another version check.

src/builder-utils.h
bbhtt: i can change it to appstream_has_version the name was a copy from flatpak_version_check above.

tests/test-builder.sh
swick: Fair enough

PR #655, opened 2025-09-01
Record licensing information of modules

swick:
builder-module: Support recording the LICENSE information of modules
    
    We want to be able to record the licensing information about the modules
    that we build in /app/share/licenses/$MODULE_NAME.
    
    The changes here do two things:
    
    1. Every module that gets build is searched for a bunch of common names
       for licensing files and every match is copied to the target dir.
    2. The new `license-files` key can be used to specify the license files
       of a module which must be found for the stage to succeed.

### DIFF
--- a/tests/test-builder.sh
+++ b/tests/test-builder.sh
@@ -61,6 +61,7 @@ cp $(dirname $0)/module2.yaml include1/include2/
 cp $(dirname $0)/source2.json include1/include2/
 cp $(dirname $0)/data2 include1/include2/
 cp $(dirname $0)/data2.patch include1/include2/
+echo "MY LICENSE" > ./LICENSE
 
 for MANIFEST in test.json test.yaml test-rename.json test-rename-appdata.json ; do
     echo "building manifest $MANIFEST" >&2
@@ -95,6 +96,8 @@ for MANIFEST in test.json test.yaml test-rename.json test-rename-appdata.json ; 
 
     ${FLATPAK} build appdir /app/bin/hello2.sh > hello_out2
     assert_file_has_content hello_out2 '^Hello world2, from a sandbox$'
+
+    assert_file_has_content appdir/files/share/licenses/org.test.Hello2/test/LICENSE '^MY LICENSE$'
 
     echo "ok build"
 done
--- a/tests/test-rename-appdata.json
+++ b/tests/test-rename-appdata.json
@@ -74,6 +74,10 @@
                             "path": "Hello.xml"
                         },
                         {
+                            "type": "file",
+                            "path": "LICENSE"
+                        },
+                        {
                             "type": "script",
                             "dest-filename": "hello2.sh",
                             "commands": [ "echo \"Hello world2, from a sandbox\"" ]
--- a/tests/test-rename.json
+++ b/tests/test-rename.json
@@ -75,6 +75,10 @@
                             "path": "Hello.xml"
                         },
                         {
+                            "type": "file",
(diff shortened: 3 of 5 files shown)


src/builder-module.c
```diff
@@ -1501,6 +1522,183 @@ find_file_with_extension (GFile *dir,
...
+                                 GFile          *source_dir,
+                                 GFile          *app_dir,
+                                 GError        **error)
+{
+  g_autoptr(GPtrArray) license_files = NULL;
+  g_autoptr(GFile) license_dir = NULL;
+
+  license_dir = g_file_new_build_filename (g_file_get_path (app_dir),
+                                           "files/share/licenses",
+                                           self->name,
```
bbhtt: can this be namespaced based on the ID? `files/share/licenses/$app_id/$module_name` That's what we ended up doing for runtimes and also what kde is using.

src/builder-module.c
```diff
@@ -1501,6 +1522,183 @@ find_file_with_extension (GFile *dir,
...
+  if (!flatpak_mkdir_p (license_dir, NULL, error))
+    return FALSE;
+
+  if (!find_license_files (self, source_dir, &license_files, error))
+    return FALSE;
+
+  for (size_t i = 0; i < license_files->len; i++)
+    {
+      GFile *license_file = license_files->pdata[i];
+      g_autofree char *dst_filename = g_strdup_printf ("LICENSE_%zu", i + 1);
```
bbhtt: Probably better to not change the dest filename and try installing it as is first? Sometimes the filename holds some meaning licence spdx code, in case of Linux firmware it is named after each of the modules etc.

bbhtt:
Would be nice to have the doc bits as well.

src/builder-module.c
```diff
@@ -1501,6 +1522,183 @@ find_file_with_extension (GFile *dir,
...
+      g_autoptr(GFile) license_file =
+        g_file_resolve_relative_path (source_dir, license_file_path);
+      g_autofree char *rel_path = NULL;
+      GFileType file_type;
+
+      rel_path = g_file_get_relative_path (source_dir, license_file);
+      if (rel_path == NULL)
+        {
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                       "Bad license file path");
```
bbhtt: Maybe this should say "outside source directory" to be clearer.

src/builder-module.c
```diff
@@ -1502,12 +1523,220 @@ find_file_with_extension (GFile *dir,
...
+  license_dir = g_file_new_build_filename (g_file_get_path (app_dir),
+                                           "files/share/licenses",
+                                           id,
+                                           self->name,
+                                           NULL);
+  if (!flatpak_mkdir_p (license_dir, NULL, error))
+    return FALSE;
+
+  if (!find_license_files (self, source_dir, &license_files, error))
+    return FALSE;
```
bbhtt: i'd reorder this a bit, otherwise it is creating a bunch of empty directories with nothing in them when there is nothing to install.

bbhtt: i think it happened to me when the module to build is already cached so source dir ends up empty except the post install stage, but I think it makes sense to verify that it had something to install before creating the directory

bbhtt:
Seems to work otherwise from my tests.

src/builder-module.c
swick: Sure, changed it.

AdrianVovk (approved):
LGTM too!

PR #658, opened 2025-09-03
builder-git: Checkout LFS objects

bbhtt:
Rebased, reworded version of https://github.com/flatpak/flatpak-builder/pull/498 and an extra commit to disable the new behaviour.


src/builder-git.c
```diff
@@ -578,6 +578,16 @@ builder_git_mirror_repo (const char     *repo_location,
                     origin, full_ref_mapping, NULL))
             return FALSE;
 
+          g_print ("Fetching LFS assets\n");
+          if (!git (mirror_dir, NULL, 0, error,
+                    "lfs", "fetch", "--all", NULL))
+            {
+              git (mirror_dir, NULL, 0, error,
+                   "lfs", "logs", "last", NULL);
```
swick: This potentially passes a set error as argument to the second `git` call. Same below.

src/builder-git.c
```diff
@@ -582,7 +582,9 @@ builder_git_mirror_repo (const char     *repo_location,
           if (!git (mirror_dir, NULL, 0, error,
                     "lfs", "fetch", "--all", NULL))
             {
-              git (mirror_dir, NULL, 0, error,
+              g_autoptr(GError) log_error = NULL;
```
swick: Ahh, this really should be squashed into the previous commit; but since you don't care about the error here you can just pass NULL to the `git lfs logs` call.

src/builder-git.c
```diff
@@ -731,14 +763,17 @@ builder_git_shallow_mirror_ref (const char     *repo_location,
             "fetch", "--depth", "1", "origin", full_ref_colon_full_ref, NULL))
     return FALSE;
 
-  if (!git (mirror_dir, NULL, 0, error,
-            "lfs", "fetch", "origin", full_ref, NULL))
+  if (git_lfs_is_available ())
     {
-      g_autoptr(GError) log_error = NULL;
+      if (!git (mirror_dir, NULL, 0, error,
```
swick: This is basically just `if (git_lfs_is_available () && !git (...))`. Removes more nesting; could be done everywhere in this commit.

src/builder-git.c
```diff
@@ -148,6 +148,31 @@ git_has_version (int major,
   return TRUE;
 }
 
+static gboolean
+git_lfs_is_available (void)
```
swick, line 175: Would make more sense to move this to the first commit and squash the checks below into the commit that actually calls git lfs. Then every commit is kept in working condition.

src/builder-git.c
```diff
@@ -476,7 +476,8 @@ builder_git_mirror_repo (const char     *repo_location,
   gboolean update = (flags & FLATPAK_GIT_MIRROR_FLAGS_UPDATE) != 0;
   gboolean disable_fsck = (flags & FLATPAK_GIT_MIRROR_FLAGS_DISABLE_FSCK) != 0;
 
-  gboolean disable_lfs = !git_lfs_is_available ();
+  gboolean disable_lfs = (flags & FLATPAK_GIT_MIRROR_FLAGS_DISABLE_LFS) != 0 ||
+                         !git_lfs_is_available ();
```
swick: Given that you check both things everywhere, this is a good candidate to pull in its own function. `builder_git_use_lfs`?

.github/workflows/check.yml
```diff
@@ -25,7 +25,8 @@ jobs:
         libjson-glib-dev shared-mime-info desktop-file-utils libpolkit-agent-1-dev libpolkit-gobject-1-dev \
         libseccomp-dev libsystemd-dev libxml2-utils libgpgme11-dev gobject-introspection \
         libgirepository1.0-dev libappstream-dev libdconf-dev clang socat flatpak \
-        libcurl4-gnutls-dev libflatpak-dev libyaml-dev elfutils git patch libarchive-tools
+        libcurl4-gnutls-dev libflatpak-dev libyaml-dev elfutils git patch libarchive-tools \
+        git-lfs
```
swick: This also should happen before (or in) the commit, that makes use of it.

src/builder-git.c
bbhtt: Yea I was debating whether to squash it to bart's commit.

src/builder-git.c
bbhtt: So just this function as the first commit?

The first commit is working, but it would fail if git lfs is not installed botching normal git operations too. I didn't want that.

src/builder-git.c
bbhtt: Makes sense.

src/builder-git.c
```diff
@@ -148,6 +148,31 @@ git_has_version (int major,
   return TRUE;
 }
 
+static gboolean
+git_lfs_is_available (void)
+{
+  static gboolean checked = FALSE;
+  static gboolean available = FALSE;
```
swick: https://docs.gtk.org/glib/type_func.Once.init_enter.html

src/builder-git.c
swick: Can add the CI bits in the first commit, then add the new --disable-lfs option in the next commit but not hook it up yet, then either add the `builder_git_use_lfs` in one commit or merge it with the rest of the changes.

src/builder-git.c
```diff
@@ -148,6 +148,83 @@ git_has_version (int major,
   return TRUE;
 }
 
+static gboolean
+git_lfs (GFile   *dir,
+         char   **output,
+         GSubprocessFlags flags,
+         GError **error,
+         ...)
```
swick: I would just take a null terminated `const char **` here and leave the vaargs to builder_git_run_lfs

src/builder-git.c
```diff
@@ -606,6 +688,11 @@ builder_git_mirror_repo (const char     *repo_location,
                     was_shallow ? "--unshallow" : NULL,
                     NULL))
             return FALSE;
+
+          g_print ("Fetching LFS assets\n");
```
swick: Well, you might not. Makes me wonder if it is really a good idea to just ignore `builder_git_run_lfs` calls when lfs is not available.

bbhtt: I think it makes sense, git sources may not have any lfs assets to fetch and the user may not have any use for lfs or lfs installed. It is being run always without that.

I can just move the messages to builder_run_git_lfs and create something generic, when they actually get executed (instead of being skipped due to not having LFS or the flags).

bbhtt: The other solution is I guess I have to add git-lfs to meson etc. to signal we want it at run time. It is not common to have git lfs installed.

It'd break fetching regular git sources too otherwise.

src/builder-git.c
bbhtt: changed in last commit, will squash it later to first one.

src/builder-git.c
bbhtt: Another option is I guess to add a global `--disable-lfs` cli arg, instead of checking for lfs availability. Those who don't have it can use that to disable for entire manifest.

src/builder-git.c
```diff
@@ -148,6 +148,94 @@ git_has_version (int major,
...
 }
 
+static gboolean
+git_lfs (GFile   *dir,
+         char   **output,
+         GSubprocessFlags flags,
+         GError **error,
+         const char **args)
+{
+  GPtrArray *full_args;
```
hfiguiere: you should use `g_autoptr()` here.

src/builder-git.c
```diff
@@ -148,6 +148,94 @@ git_has_version (int major,
...
+
+static gboolean
+builder_git_run_lfs (GFile       *dir,
+                     FlatpakGitMirrorFlags flags,
+                     GError     **error,
+                     const char  *subcommand,
+                     ...)
+{
+  va_list ap;
+  GPtrArray *args;
```
hfiguiere: also here user `g_autoptr()`

src/builder-git.c
bbhtt: changed,

swick (approved):

src/builder-git.c
```diff
@@ -858,10 +953,18 @@ builder_git_checkout (const char     *repo_location,
             "config", "--bool", "core.bare", "false", NULL))
     return FALSE;
 
+  if (!builder_git_run_lfs (dest, mirror_flags, error,
+                            "install", NULL))
+    return FALSE;
+
```
bbhtt: hmm, i wonder if this should use `git lfs install --skip-smudge`

bbhtt: lfs activates the smudge filter by default, so the checkout after that (line 960-961) is pulling the lfs files anyway even with `disable-lfs`, since we are explicitly pulling LFS this should probably use `--skip-smudge`.

bbhtt: actually, the git lfs install step might not be needed at all as we are explicitly fetching and checking out.

bbhtt: ok,

the `git lfs install` part in the first commit was wrong, as it was affecting global git config `~/.gitconfig` instead of per repo git config when it is running inside `.flatpak-builder/build/foobar-1`. The correct one to use would have been `git lfs install --local`.


`git checkout` will try to pull lfs files anyway in presence of a `/etc/gitconfig, ~/.gitconfig` with lfs filters. if the user explicitly asked us to disable lfs on a source, `git checkout` should not pull LFS at all for that. So we need to do the counterintuitive `git lfs install --local --skip-smudge --skip-repo` for that source which tells lfs to not install hooks and install the smudge filter but with `--skip` so that a plain git checkout does not try to pull LFS files.

bbhtt: Changed it in the latest version and added a comment.

bbhtt: Hmm, in this case it causes LFS to become a dependency to disable fetching LFS objects. I guess it might be better to assume in presence of `filter lfs` in `/etc/gitconfig, ~/.gitconfig,...` the user wants to always pull LFS and probably better to handle disabling that outside flatpak-builder.

bbhtt: > the git lfs install part in the first commit was wrong, as it was affecting global git config ~/.gitconfig instead of per repo git config when it is running inside .flatpak-builder/build/foobar-1. The correct one to use would have been git lfs install --local.

Changed it back to the original with this fix. 

I will add a note in the doc instead.

bbhtt:
Deployed this to Flathub to test on https://github.com/flathub/dev.mariinkys.Oboete/pull/15 s

Seems to work.

Combined with this change, https://github.com/flathub/org.flatpak.Builder/commit/1b28ed64ff2d700d915903789df7cf2c67b89eb9

Issue #594, opened 2024-03-25
Have CI that builds with --wrap-mode=nodownload


nanonyme:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

Currently this project builds with submodules. But when submodules are broken, Meson automatically switches to wraps. Let's have two CI modes from Meson
1. submodules are *disabled*. We test that wraps work
2. submodules are *enabled* but --wrap-mode=nodownload is passed. if submodules don't work, then build fails

nanonyme:
@TingPing this might be material for after autotools is removed

Issue #593, opened 2024-03-22
Latest flatpak-builder tag doesn't build with libglnx submodule


nanonyme:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a bug that matches the one I want to file, without success.

### flatpak-builder version

1.4.2

### Flatpak version

1.15.6

### How to reproduce

Build flatpak-builder 1.4.2 tag with libglnx submodule (no wrap)

### Expected Behavior

Build succeeds

### Actual Behavior

Build fails because 1.4.2 (accidentally?) switched libglnx submodule to 5 years old version that doesn't support Meson

### Additional Information

This was already fixed through https://github.com/flatpak/flatpak-builder/commit/ae86bde81fbb0316f08121793c0824c18b1cfb9a but there is no tag containing the fix.

TingPing:
Fixed in 1.4.3 release.

PR #659, opened 2025-09-03
ci: Add a wrap only build mode

bbhtt:
Fixes https://github.com/flatpak/flatpak-builder/issues/594

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -166,3 +166,9 @@ jobs:
         path: |
           builddir/meson-logs/testlog.txt
           installed-test-logs/
+
+    - name: Configure flatpak-builder with Meson wraps
+      run: meson setup --wrap-mode=forcefallback ${BUILDDIR}_wrap .
+
+    - name: Build flatpak-builder with Meson wraps
+      run: meson compile -C ${BUILDDIR}_wrap


bbhtt:
See https://github.com/flatpak/flatpak-builder/issues/593 for context.

.github/workflows/check.yml
```diff
@@ -138,7 +138,7 @@ jobs:
 
     - name: Configure flatpak-builder
       # TODO: Enable gtk-doc builds
-      run: meson . ${BUILDDIR}
+      run: meson setup . ${BUILDDIR}
```
swick: I didn't even know it can take two args, but my meson tells me it's `meson setup ... [builddir] [sourcedir]`.

bbhtt: Yea looks like it was switched at some point.

.github/workflows/check.yml
```diff
@@ -154,3 +154,9 @@ jobs:
         path: |
           builddir/meson-logs/testlog.txt
           installed-test-logs/
+
+    - name: Configure flatpak-builder with Meson wraps
+      run: meson setup --wrap-mode=forcefallback . ${BUILDDIR}_wrap
```
swick: Still using `[sourcedir] [builddir]`

bbhtt: missed the second one thanks

swick (approved):

PR #665, opened 2025-09-11
tests: Add a way to skip the test involving mirroring screenshots

smcv:
Some build and CI systems operate fully offline in order to avoid dependencies on remote services. Skip this test if environment variable FLATPAK_BUILDER_TESTS_OFFLINE is set non-empty.


bbhtt:
I went with https://github.com/flatpak/flatpak-builder/pull/666 instead to keep the test working even without network, but I may use the variable idea to guard future tests that may need network.

PR #666, opened 2025-09-11
tests: Allow screenshot ref commit test to run offline

bbhtt:
This makes it harder for me to write the test for https://github.com/flatpak/flatpak-builder/pull/652, but if offline tests are wanted, so be it.

I didn't delete the test files, as I plan to reuse them later for the test in #652. I need some fully qualified and valid desktop file, metainfo file for that.


bbhtt:
Any feedback on this? We probably should include this in the next release.

swick:
Looks fine to me. Making things work without invoking the network is always good IMO.

bbhtt:
OK let's go with this.

Issue #341, opened 2020-04-29
Feature request:: log commits of used Sdk's and other components


nanonyme:
It would be nice if flatpak-builder logged commits at the beginning of build of things needed for making a build. This would allow you to do a quick comparison of what the Flathub server has and what you have to ensure that they are actually the same before debugging further if the build fails in unexpected ways.

PR #667, opened 2025-09-12
builder-manifest: Log sdk, runtime and base versions and commits used

bbhtt:
Fixes https://github.com/flatpak/flatpak-builder/issues/341

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -1708,6 +1708,12 @@ builder_manifest_start (BuilderManifest *self,
                          self->sdk,
                          builder_manifest_get_runtime_version (self));
 
+  if (self->sdk_commit != NULL)
+    g_debug ("Using %s version %s commit %s",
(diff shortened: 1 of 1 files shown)


src/builder-manifest.c
```diff
@@ -1708,6 +1708,12 @@ builder_manifest_start (BuilderManifest *self,
                          self->sdk,
                          builder_manifest_get_runtime_version (self));
 
+  if (self->sdk_commit != NULL)
+    g_debug ("Using %s version %s commit %s",
```
swick, line 1712: The gnome handbook says that one should use parenthesis if there are multiple lines. I know that this is not followed consistently in flatpak-related project, but I wonder if we should enforce this.

bbhtt: i don't have a strong opinion on this, but it is even inconsistent in this file. the instances i could spot except a few are not using braces.

Issue #475, opened 2022-06-10
[Bug]: flatpak-builder is happy to delete the current directory


allisonkarlitskaya:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a bug that matches the one I want to file, without success.

### flatpak-builder version

flatpak-builder 1.2.2

### Flatpak version

Flatpak 1.12.7

### How to reproduce

1. Start in the source directory of the project you're working on
2. Try to remember how to run `flatpak-builder`.  Something about `--force-clean --user --install` or so, I think?
3. See the `--help` output mentions that you need to give mandatory `DIRECTORY` and `MANIFEST` options.
4. Perhaps having used Docker a couple of times, try giving `.` for the `DIRECTORY` argument.

### Expected Behavior

A nice message from `flatpak-builder` saying "refusing to delete the current directory".

### Actual Behavior

`flatpak-builder` did exactly what I told it to.  `--force-clean .` is pretty clear, no?

So maybe this isn't a bug, but it's a big (and destructive) enough trap that's easy enough to fall into that it ought to be forbidden.  Maybe it makes sense to build in the current directory, but I can't imagine any situation in which it's reasonable to delete the current directory.  `flatpak-builder --force-clean .` should refuse to operate.

There's precedent for this: even `rm -rf` prevents you from shooting yourself in the foot in this particular way:

```
rm: refusing to remove '.' or '..' directory: skipping '.'
```

### Additional Information

Fortunately, I take automated hourly snapshots of my `~/src`.  Combined with some `git show` output that happened to be in the terminal backlog I was able to piece everything back together.  This time.  :sweat_smile:

PR #668, opened 2025-09-12
builder-main: Prevent deleting state-dir, cwd or their parents with force-clean

bbhtt:
Fixes https://github.com/flatpak/flatpak-builder/issues/475

I guess the question here is how much mistake should it prevent and how much is this a breaking change.

I think the spinoff state dir deletion case that I found makes sense and wouldn't really be a breaking change as that never worked without hanging. That also automatically solves the builddir being `.` case at least when `--state-dir` doesn't point to somewhere else.

I'm a bit unsure about the blocking the cwd case.

### DIFF
--- a/doc/flatpak-builder.xml
+++ b/doc/flatpak-builder.xml
@@ -474,7 +474,9 @@
 
                 <listitem><para>
                     Erase the previous contents of DIRECTORY if it is
-                    not empty.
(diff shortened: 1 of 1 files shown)


bbhtt:
After thinking about it for a while, I think it's fine for 1.4.6. I couldn't find a use case which cannot be easily solved by adjusting `DIRECTORY`.

I updated the doc.

PR #672, opened 2025-09-19
ci: Specify permissions

bbhtt:
There are too many jobs now let's make sure to reduce duplicate runs.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -12,6 +12,9 @@ env:
   DEBIAN_FRONTEND: noninteractive
   TESTS_TIMEOUT: 10 # in minutes
 
+permissions:
+  contents: read
+
 jobs:
   check-autotools:
     name: Build with gcc and test


.github/workflows/check.yml
```diff
@@ -16,6 +16,8 @@ jobs:
   check-autotools:
     name: Build with gcc and test
     runs-on: ubuntu-22.04
+    permissions:
```
swick: Should be able to set that at the top level instead of per-job, no?

.github/workflows/check.yml
```diff
@@ -16,6 +16,9 @@ jobs:
   check-autotools:
     name: Build with gcc and test
     runs-on: ubuntu-22.04
+    concurrency:
+      group: ci-gcc-${{ github.ref }}
```
swick: Mh, I don't think I understand what problem this is solving. Can you expand on that?

.github/workflows/check.yml
bbhtt: In this case yes, but we have to make it per job again if a certain step requires additional permissions.

.github/workflows/check.yml
bbhtt: It makes redundant runs auto cancel. Try pushing to a branch/PR two or more times you will have multiple duplicate copies of the same jobs running for each push.

Only the latest pipeline from latest push is required, old pipelines can be auto cancelled.

.github/workflows/check.yml
swick: Right, but from the principle of least privilege we should limit the workflow to only what is needed and then at the job level refine it further. So currently we only need read permission, so the workflow should only run with read permission.

.github/workflows/check.yml
swick: Well, they are not duplicate because they run for different commits. Not sure if this is really desirable, but if you want it, let's just go with it.

.github/workflows/check.yml
bbhtt: > Right, but from the principle of least privilege

you might have a job that requires no permissions as well. adding it to the toplevel applies the permission to that as well.

although that's rare, content: read is for checkout action and almost any job will need it.

.github/workflows/check.yml
bbhtt: they are duplicate in the sense you want to know the pipeline status from the latest push only
the old pipelines will be running for older commits and may fail for something you already have fixed in the latest commit.

.github/workflows/check.yml
swick: > you might have a job that requires no permissions as well

Yes? Then you specify that at the job level. The point would be to cut down the permissions for the entire workflow as much as possible in the first place.

.github/workflows/check.yml
swick: Like I said, not entirely convinced. If I push, I usually want to see the result of that, even if I add another commit while the CI is still running for the previous one.

bbhtt: ok i dropped this, this is what i usually use.

swick: If you feel strongly about it, then please do add it back.

bbhtt: I can live without it, it's nice to have, not essential as the jobs don't take long.

Issue #476, opened 2022-06-14
[Bug]: `cleanup` directive in manifest does not delete directories ending with `/`


Yvan-Masson:
### Checklist

- [X] I agree to follow the [Code of Conduct](https://github.com/flatpak/flatpak/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a bug that matches the one I want to file, without success.

### flatpak-builder version

1.2.2

### Flatpak version

1.12.7

### How to reproduce

Take a sample manifest like this one (which is missing here required dependencies to speed up build):
```
app-id: fr.masson_informatique.WhosWho
runtime: org.gnome.Platform
runtime-version: '42'
sdk: org.gnome.Sdk
command: whoswho
cleanup:
  - "/bin/"
finish-args:
  - --share=ipc
  - --socket=fallback-x11
  - --socket=wayland
  - --talk-name=org.gtk.vfs.*
  - --filesystem=xdg-run/gvfsd

modules:
  - name: WhosWho
    buildsystem: simple
    build-commands:
      - ./build-translations.py
      - pip3 install --prefix=${FLATPAK_DEST} --no-deps .
    sources:
      - type: archive
        url: https://framagit.org/Yvan-Masson/WhosWho/-/archive/v1.4/WhosWho-v1.4.tar.gz
        sha256: e022b5053c41cf3e70d78ccc695983a5a1bfebf696b4d54a3dad4a2d02df8199
```
Note that removing `/bin` in `cleanup` is again just for the test.

### Expected Behavior

Directory `/bin` should be deleted.

### Actual Behavior

In this case `/bin/` is not deleted. However, if you remove the trailing `/`, it is.

PR #677, opened 2025-09-23
builder-flatpak-utils: Normalize the single trailing slash in pattern

bbhtt:
Fixes https://github.com/flatpak/flatpak-builder/issues/476

### DIFF
--- a/src/builder-flatpak-utils.c
+++ b/src/builder-flatpak-utils.c
@@ -156,6 +156,20 @@ flatpak_path_match_prefix (const char *pattern,
             return string;
           return NULL;
 
+        case '/':
+          if (*pattern == 0)
+            {
+              /* treat as directory boundary, don't consume match */
+              if (*string == '/' || *string == 0)
+                return string;
+              return NULL;
+            }
+          if (*string == '/')
+            string++;
+          else
+            return NULL;
+          break;
+
         case '?':
           if (*string == '/' || *string == 0)
             return NULL;
--- a/tests/test-builder-cleanup.sh
+++ b/tests/test-builder-cleanup.sh
@@ -23,7 +23,7 @@ set -euo pipefail
 
 skip_without_fuse
 
-echo "1..8"
+echo "1..9"
 
 setup_repo
 install_repo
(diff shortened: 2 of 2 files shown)


bbhtt:
any comments on this?

src/builder-flatpak-utils.c
swick, line 157: This handles the inverse. Could add something like

```c
case '/':
  if (*string == '/')
    string++;
  else if (*string != 0)
    return NULL;
  break;
```

swick:
Not sure if I would consider this a bug in the first place. Kind of depends on what you consider a path prefix I guess.

bbhtt:
according to the man, patterns are taken to be either the full pathname or basename, not normalising is a bit weird. it also accepts `/*`.

src/builder-flatpak-utils.c
bbhtt: that does not work directly

bbhtt: i had to do this, but this copies case 0 again.

```
case '/':
  if (*pattern == 0)
    {
      /* treat as directory boundary, don't consume */
      if (*string == '/' || *string == 0)
        return string;
      return NULL;
    }
  if (*string == '/')
    string++;
  else if (*string != 0)
    return NULL;
  break;
```

src/builder-flatpak-utils.c
```diff
@@ -156,6 +156,20 @@ flatpak_path_match_prefix (const char *pattern,
...
+            {
+              /* treat as directory boundary, don't consume match */
+              if (*string == '/' || *string == 0)
+                return string;
+              return NULL;
+            }
+          if (*string == '/')
+            string++;
+          else if (*string != 0)
+            return NULL;
```
swick, line 170: I'm confused about this part. If the pattern is `a/b` and the match string is `a` then `*pattern != 0`, and `*string == 0` and we continue matching the string against `b`.

IOW, shouldn't this be

```c

          if (*string == '/')
            string++;
          else
            return NULL;
```

bbhtt: yes this seems like a bug, if the match string is exhausted it should return NULL immediately. 

this makes the cleanup greedy too against wrong patterns, I will add a test for this.

PR #690, opened 2025-10-20
Convert libglnx to subtree

bbhtt:


bbhtt:
We can update but the commits will squashed, here it is with squashed adding https://github.com/flatpak/flatpak-builder/pull/692

PR #692, opened 2025-10-21
Convert libglnx to subtree

bbhtt:


bbhtt:
This is at same commit added with:

```
git subtree add --prefix=subprojects/libglnx libglnx 202b294e6079e23242e65e0426f8639841d1210b --squash
```

Updating looks like this:

```
git subtree pull --prefix=subprojects/libglnx libglnx master --squash
```


image

PR #696, opened 2025-11-27
builder-manifest: Override branch with --default-branch when specified

alatiera:
builder-manifest: Override branch with --default-branch when specified

Commit 00f63cf added default-branch with the stated intention of being
able to override the branch specified in the manifest. The name is a bit
confusing.

However the scenario tested was only when there was no
specified branch at all, and thus the behavior was a bit buggy.

Till now, if you had both specified "branch": "something" in the
manifest and used --default-branch=test, the end result in the
ostree repo would be for flatpak-builder to use the branch name
specified in the manifest.

```
$ flatpak-builder --default-branch=test --force-clean --repo=_build/repo/ app/ tests/test-default-branch.json
$ ostree refs --repo=_build/repo/
app/org.test.DefaultBranch/x86_64/something
```

With this change, we properly pick up default-branch if its set,
since it can only be set through the cli and its the explicit intention
that we want to override it.

```
$ meson devenv -C _build/ flatpak-builder  --default-branch=test --user --force-clean --repo=repo_test/ app/ ../tests/test-default-branch.json
$ ostree refs --repo=_build/repo_test
app/org.test.DefaultBranch/x86_64/test
```

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -1477,19 +1477,25 @@ builder_manifest_get_runtime_version (BuilderManifest *self)
   return self->runtime_version ? self->runtime_version : "master";
 }
 
+/* default-branch can only be set through the cli, where branch
+ * is extracted from the manifest. As such the intended behavior
+ * is that the cli arguments should always override the branch
+ * set in the manifest, so your manifest can set the branch
+ * but you are still able to do --default-branch=test-build
+ */
 const char *
 builder_manifest_get_branch (BuilderManifest *self,
                              BuilderContext  *context)
 {
-  if (self->branch)
-    return self->branch;
-
   if (context &&
       builder_context_get_default_branch (context))
     return builder_context_get_default_branch (context);
 
   if (self->default_branch)
     return self->default_branch;
+
+  if (self->branch)
+    return self->branch;
 
   return "master";
 }
--- a/tests/test-default-branch.json
+++ b/tests/test-default-branch.json
@@ -1,0 +1,26 @@
+{
+    "id": "org.test.DefaultBranch",
+    "runtime": "org.gnome.Platform",
+    "sdk": "org.gnome.Sdk",
+    "command": "hello",
+    "branch": "something",
+    "modules": [
+        {
+            "name": "install_test",
+            "buildsystem": "simple",
+            "build-commands": [
+                "mkdir -p ${FLATPAK_DEST}/bin",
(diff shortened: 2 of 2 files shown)


tests/test-default-branch.json
```diff
@@ -0,0 +1,26 @@
+{
+    "id": "org.test.DefaultBranch",
+    "runtime": "org.gnome.Platform",
```
alatiera, line 3:
```suggestion
    "runtime": "org.test.Platform",
```

Didn't want to bother figuring out where the test runtime was

alatiera:
I also want to write proper tests for this before we merge it.

The scenarios we need to test are:

* branch not specified -> master
* branch not specified && --default-branch -> default-branch
* branch specified -> specified
* branch specified  && --default-branch -> default-branch

alatiera:
lol I never wrote the commit subject

src/builder-manifest.c
```diff
@@ -1477,20 +1477,26 @@ builder_manifest_get_runtime_version (BuilderManifest *self)
   return self->runtime_version ? self->runtime_version : "master";
 }
 
+/* default-branch can only be set through the cli, where branch
```
Erick555, line 1480: default-branch can be set through [manifest](https://github.com/flathub/com.valvesoftware.Steam.CompatibilityTool.Proton-GE/blob/20d2c86b7ebaa1e265becfa16b936dd0d3e66628/com.valvesoftware.Steam.CompatibilityTool.Proton.yml#L2) though.

bbhtt:
It was intentional that the `branch` key in the manifest overrides both the `default-branch` key and the `--default-branch` argument.

I don't agree with the complicated design; so my plan was to remove the `branch` key entirely and just have the default-branch key+arg where the arg overrides the manifest key like everything else.

But I guess making both being overridden by the arg is ok.

In any case this is a breaking change and has to go through unstable branch first.

This also needs coordination at least on Flathub before merging as there are a number of extensions deriving their expected branch from the `branch` manifest key.

alatiera:
Indeed, re-reading the manifest it actually works as expected.

> This key overrides both the default-branch key, and the --default-branch commandline option. 

Where this goes wrong imo is the following then.

* You can't specify `--branch` but you can specify `--default-branch`.
* If you already have a `branch` set, `--default-branch` should have been erroring out since it wouldn't take effect.
* Initially we only had `branch` so all our old manifests still use that.

Let's close this PR and instead go fix the manifests then I guess

PR #706, opened 2026-03-07
builder-module: Record licence files in licences directory as well

bbhtt:
The licences directory is a common convention across Qt and KDE
projects. See examples [1], [2], [3]. The files inside that directory
usually use the licence id as their filename. Listing them individually
inside licence-files key is annoying but without anything no licence
information is recorded for these projects. So when collecting the
default license files, also collect files inside the licences directory.

[1]: https://github.com/qt/qtbase/tree/2d790125429123aec4fb3dbee8335732f2d122eb/LICENSES
[2]: https://github.com/qt/qtwebengine/tree/cf5db2b9242facf4797a7779ae50916326f75d0b/LICENSES
[3]: https://invent.kde.org/system/dolphin/-/tree/02165defeeeb529a566c8cc00f30e7a5c8c11185/LICENSES

### DIFF
--- a/src/builder-module.c
+++ b/src/builder-module.c
@@ -87,6 +87,18 @@ typedef struct {
   char *dst_name;
 } LicenseFile;
 
+typedef struct
+{
+  const char * const *patterns;
+  const char         *prefix;
+  GPtrArray          *files;
+} LicenseScanData;
+
+typedef gboolean (*LicenseScanCb) (int         fd,
+                                   const char *name,
+                                   gpointer    user_data,
+                                   GError    **error);
+
 static void serializable_iface_init (JsonSerializableIface *serializable_iface);
 
 G_DEFINE_TYPE_WITH_CODE (BuilderModule, builder_module, G_TYPE_OBJECT,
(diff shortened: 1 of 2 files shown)


src/builder-module.c
```diff
@@ -1624,6 +1632,43 @@ find_default_license_files (BuilderModule  *self,
...
+          g_autoptr(GFile) license_file = NULL;
+
+          if (g_file_info_get_file_type (sub_info) != G_FILE_TYPE_REGULAR)
+            continue;
+
+          license_file = g_file_enumerator_get_child (subdir_enum, sub_info);
+          g_ptr_array_add (files, g_steal_pointer (&license_file));
+        }
+
+      if (subdir_error != NULL)
```
swick: Wondering about the error handling here a bit because this early return means we won't always try to find all the files if an error occurs in one subdir.

bbhtt: changed it to skip and warn, i don't find failures in the default case.

swick: Looking at this again, I think the right way to handle errors is to bail out as early as possible. The loop above does it as well. The code here just ignores a bunch of errors and now warns instead of erroring out. Maybe it makes sense to combine both loops (we iterate through all files anyway) which should reduce the error conditions and then early exit as soon as the subdir iter goes wrong?

src/builder-module.c
```diff
@@ -1727,7 +1735,30 @@ find_default_license_files (BuilderModule  *self,
...
+  if (!scan_license_dir (source_dfd, NULL, TRUE, files, error))
+    return FALSE;
+
+  for (size_t i = 0; license_dir_names[i] != NULL; i++)
+    {
+      glnx_autofd int sub_dfd = -1;
+
+      if (!glnx_opendirat (source_dfd, license_dir_names[i], FALSE, &sub_dfd, error))
+        {
+          if (*error && (*error)->domain == G_IO_ERROR &&
```
swick: `error` is an optional in parameter. Use a `local_error` here and propagate it.

swick: would be easier to read if this was just two `g_error_matches`

src/builder-module.c
```diff
@@ -1727,7 +1735,30 @@ find_default_license_files (BuilderModule  *self,
...
+              G_IN_SET ((*error)->code, G_IO_ERROR_NOT_FOUND, G_IO_ERROR_NOT_DIRECTORY))
+            {
+              g_clear_error (error);
```
swick: Another rule is that out params (which error is) should only be modified once on success. But this again would be solved by using a local error.

src/builder-module.c
```diff
@@ -1625,6 +1625,7 @@ static const char *default_licence_file_patterns[] = {
 static gboolean
 scan_license_dir (int          dirfd,
                   const char  *prefix,
+                  gboolean     match_licence_file_pattern,
```
swick: Not a fan of boolean arguments, and specifically in this case the function does very different things depending on it. Maybe a callback would be more appropriate here?

src/builder-module.c
```diff
@@ -1624,6 +1624,7 @@ static const char *default_licence_file_patterns[] = {
 
 static gboolean
 scan_license_dir (int          dirfd,
+                  const char  *prefix,
```
swick: If you go for a callback, you could even pass in the prefix as `gpointer userdata` and remove that bit of logic as well.

bbhtt:
changed it to use callbacks, it's easily extensible that way for future but i don't think this needs more licence file handling for a while.

src/builder-module.c
```diff
@@ -1726,6 +1726,35 @@ license_scan_cb (int         fd,
                                error);
 }
 
+static gboolean
+scan_license_subdir (int               parent_dfd,
```
swick, line 1730: I would inline this function into `find_default_license_files`, but this is getting very nitpicky.

swick (approved):

src/builder-module.c
bbhtt: It's outside incase I need to add more subdirectories later on.

PR #711, opened 2026-03-08
Support installing files from file sources

### LINKED ISSUE #54: Provide a more packager-friendly way of installing appdata files
The thing is if an appdata file is missing, one has to add it as a source and then either modify `build-commands` or `post-install` and add something like `"mkdir -p /app/share/appdata", "install -Dm644 appdata.xml /app/share/appdata/appname.appdata.xml"` and add a `"rename-appdata-file": "appname.appdata.xml"`, but with this approach a simple appdata change requires a rebuild, and, frankly, the approach is a bit tedious.

It would be cool if there was something like:
```
      "sources": [
        {
          "type": "appdata",
          "path": "appdata.xml"
        }
      ]
```

which would perform all the above, automatically, without needing a rebuild.

bbhtt:
This adds a new install stage to BuilderSourceFile that copies files to
a specified directory within the appdir after post-install for
sources that have opted into using it (currently only the file source
and no-op for everyone else).

Additionally, the file source now has a install-dir and install-mode
property. The install-dir property assumes the install location
is relative to FLATPAK_DEST. If the destination file already exists, the
install is skipped with a warning. If install-mode is specified as a
valid octal permission string, the permissions are applied on the file
otherwise the file retains its original permissions.

Fixes: https://github.com/flatpak/flatpak-builder/issues/54

### DIFF
--- a/src/builder-source-file.c
+++ b/src/builder-source-file.c
@@ -46,6 +46,8 @@ struct BuilderSourceFile
   char         *dest_filename;
   char         *http_referer;
   gboolean      disable_http_decompression;
+  char         *install_dir;
+  char         *install_mode;
 };
 
 typedef struct
@@ -67,6 +69,8 @@ enum {
   PROP_MIRROR_URLS,
   PROP_HTTP_REFERER,
   PROP_DISABLE_HTTP_DECOMPRESSION,
+  PROP_INSTALL_DIR,
+  PROP_INSTALL_MODE,
   LAST_PROP
 };
 
@@ -84,6 +88,8 @@ builder_source_file_finalize (GObject *object)
   g_free (self->dest_filename);
   g_free (self->http_referer);
   g_strfreev (self->mirror_urls);
+  g_free (self->install_dir);
+  g_free (self->install_mode);
 
   G_OBJECT_CLASS (builder_source_file_parent_class)->finalize (object);
 }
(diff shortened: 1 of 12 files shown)


src/builder-source-file.c
```diff
@@ -552,6 +590,87 @@ builder_source_file_extract (BuilderSource  *source,
...
+    return flatpak_fail (error, "Source file '%s' not found", filename);
+
+  if (!flatpak_mkdir_p (install_dir, NULL, error))
+    return FALSE;
+
+  if (flatpak_file_query_exists_nofollow (dst))
+    g_printerr ("Warning: %s already exists, skipping install\n", dst_path);
+  else
+    {
+      if (!g_file_copy (src, dst, G_FILE_COPY_NOFOLLOW_SYMLINKS,
```
swick, line 647: AFAIU this is executed in the host context and `dst` can point to arbitrary locations on the filesystem. You really have to make sure that `dst` points somewhere inside `dest_dir`. This is harder than it sounds because even if you ensure there are no ".." in `self->install_dir`, there might be symlinks anywhere in the build-controlled path (and they *do* get followed even if you use `G_FILE_COPY_NOFOLLOW_SYMLINKS`; they only care about the last element in the path).

I think we have a very similar problem with the LICENSE file code (i.e. you can copy certain files from the host system to the build output).

The solution is the use something like openat2/chase (https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/64).

bbhtt: Yea this is indeed executed in host context.

swick: Urgh, `builder_migrate_locale_dirs` also got this wrong already. I'm wondering if people *assume* that flatpak-builder is secure to run on untrusted content, because it certainly doesn't look like it is.

PR #715, opened 2026-03-11
builder-module: Abort on invalid module names

bbhtt:
This goes a step above 6d1d9724c5b30c680e5eefe05f5ddf7b15fdc92e and makes it an actual error. The name is used to construct paths in various places


swick (approved):

swick:
We'll need to backport this to the 1.4 series, it fixes path traversal issues.

PR #732, opened 2026-04-11
builder-context: Ensure sdk config path resolution stays beneath sdk path

bbhtt:


src/builder-sdk-config.h
```diff
@@ -40,7 +40,7 @@ const char *      builder_sdk_config_get_cxxflags (BuilderSdkConfig  *self);
 const char *      builder_sdk_config_get_ldflags  (BuilderSdkConfig  *self);
 const char *      builder_sdk_config_get_rustflags  (BuilderSdkConfig  *self);
 
-BuilderSdkConfig *builder_sdk_config_from_file    (GFile             *file,
+BuilderSdkConfig *builder_sdk_config_from_fd      (int                fd,
```
swick, line 43: This commit leaves the code in a broken state because the callers don't get adjusted here.

src/builder-context.c
```diff
@@ -1183,19 +1183,35 @@ builder_context_load_sdk_config (BuilderContext   *self,
                                  const char       *sdk_path,
                                  GError          **error)
 {
-  g_autoptr(GFile) root = g_file_new_for_path (sdk_path);
-  g_autoptr(GFile) config_file = g_file_resolve_relative_path (root, "files/etc/flatpak-builder/defaults.json");
-  g_autoptr(GError) local_error = NULL;
+  glnx_autofd int root_dfd = -1;
   g_autoptr(BuilderSdkConfig) sdk_config = NULL;
+  g_autoptr(GError) local_error = NULL;
+  int fd;
```
swick: That variable is owning the fd, so use glnx_autofd. `builder_sdk_config_from_fd` does not take ownership, so it's leaking currently.

swick (commented):
Looks good in general, only two smaller issues.

src/builder-sdk-config.h
bbhtt: merged it with the next.

src/builder-context.c
bbhtt: oops, fixed.

src/builder-utils.c
```diff
@@ -583,6 +583,42 @@ builder_gobject_from_data (GType       gtype,
...
+    }
+
+  if ((fd_flags & O_PATH) != 0 ||
+      (fd_flags & O_ACCMODE) == O_WRONLY)
+    {
+      rd_fd = glnx_fd_reopen (fd, O_RDONLY, error);
+      if (rd_fd < 0)
+        return NULL;
+
+      stream = g_unix_input_stream_new (rd_fd, TRUE);
```
swick: rd_fd is owning, but this takes ownership, so you must steal the fd here, or just not take ownership which works because the steam is in the same scope as rd_fd.

Another pattern I like to use when it comes to ownership which depends on which case is taken is:
```
int stream_fd;
glnx_autofd int stream_fd_owned = -1;

...


if (foo)
  stream_fd = stream_fd_owned = glnx_fd_reopen (fd, O_RDONLY, error);
else
  stream_fd = fd;

// use stream_fd
```

swick (commented):
With that issue fixed, the change LGTM

src/builder-utils.c
bbhtt: yea that's cleaner switched it.

swick (approved):

PR #733, opened 2026-04-11
builder-utils: Rewrite locale migration to use fd-based traversal

bbhtt:


src/builder-utils.c
```diff
@@ -395,18 +442,22 @@ gboolean
...
+  if (!glnx_opendirat (AT_FDCWD,
+                       flatpak_file_get_path_cached (root_dir),
+                       FALSE, &root_dfd, error))
+    return FALSE;
+
+  if (!builder_ensure_dirs_at (root_dfd, separate_components, &separate_dfd, error))
+    return FALSE;
 
-  if (!migrate_locale_dir (lib_locale_dir, separate_dir, "lib", error))
+  if (!migrate_locale_dir (root_dfd, "lib/locale", separate_dfd, "lib", error))
```
swick: I know this is marked as draft but this "lib/locale" gets put into `glnx_opendirat` which means lib can be a symlink.

bbhtt:
I still need to test this and find more examples of how locale directories are structured by apps.

The previous code was recursively copying everything inside the localedir which isn't possible to do when this is fd based. I don't think we can support the exact same behaviour.

I managed to support the gettext localedir structure and the when locale files are inside the immediate locale directory. AppStream also mainly handles these two structures too. I don't think other structures exists but who knows what some random software buildsystem is doing.

If it can't be fully emptied, it cannot be removed or symlinked, then it gets left in a weird state and errors from unlinkat or symlinkat cannot sanely be ignored too.

src/builder-utils.c
bbhtt: this is now passed to chaseat relative to rootdfd should be fixed i think.

bbhtt:
Ok maybe this is closer to the previous behaviour, did something to recursively migrate things.

bbhtt:
Needs a rebase over other PRs, but I think it's ok now, minus me testing it some more.

bbhtt:
it would be nice to have reviews on this.

src/builder-utils.c
```diff
@@ -284,77 +284,244 @@ directory_is_empty (const char *path)
...
+  glnx_autofd int locale_subdir_dfd = -1;
+  g_autofree char *language = g_strdup (lang_name);
+  g_autofree char *target = NULL;
+
+  if (!locale_name_to_language (language, lang_name, error))
+    return FALSE;
+
+  const char *components[] = { language, subdir, lang_name, NULL };
+
+  if (!builder_ensure_dirs_at (separate_dfd, components, &locale_subdir_dfd, error))
```
swick, line 392: Would be nice to replace this with https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/76.

src/builder-utils.c
```diff
@@ -284,77 +284,244 @@ directory_is_empty (const char *path)
...
+      if (!glnx_dirfd_iterator_next_dent (&iter, &dent, NULL, error))
+        return FALSE;
+
+      if (dent == NULL)
+        break;
+
+      g_ptr_array_add (names, g_strdup (dent->d_name));
+    }
+
+  for (size_t i = 0; i < names->len; i++)
```
swick: Not sure why this is split into two loops?

src/builder-utils.c
```diff
@@ -284,77 +284,244 @@ directory_is_empty (const char *path)
...
+        if (!locale_name_to_language (language, name, error))
+          return FALSE;
+
+        /* We ship english and C locales always */
+        if (strcmp (language, "C") == 0 ||
+            strcmp (language, "en") == 0)
+          continue;
+      }
+
+      glnx_gen_temp_name (tmp_name);
```
swick: Not really sure why the rename to a tmp name is needed.

src/builder-utils.c
```diff
@@ -284,77 +284,244 @@ directory_is_empty (const char *path)
...
+
+      if (!glnx_renameat (source_dfd, name, source_dfd, tmp_name, error))
+        return FALSE;
+
+      lang_dfd = glnx_chaseat (source_dfd, tmp_name,
+                               GLNX_CHASE_RESOLVE_BENEATH |
+                               GLNX_CHASE_MUST_BE_DIRECTORY,
+                               error);
```
swick: Get the fd before, a rename does not change the inode and this not the fd.

src/builder-utils.c
```diff
@@ -284,77 +284,244 @@ directory_is_empty (const char *path)
...
+            {
+              g_clear_error (error);
+              if (!glnx_renameat (source_dfd, tmp_name, source_dfd, name, error))
+                return FALSE;
+              continue;
+            }
+          return FALSE;
+        }
+
+      if (separate_dfd < 0 &&
```
swick: Seems wrong to have this in the loop when it is invariant.

src/builder-utils.c
```diff
@@ -284,77 +284,244 @@ directory_is_empty (const char *path)
...
-                                        G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS,
-                                        NULL, NULL);
-  if (!dir_enum)
-    return TRUE;
+  if ((c = strchr (name, '@'))) *c = '\0';
+  if ((c = strchr (name, '_'))) *c = '\0';
+  if ((c = strchr (name, '.'))) *c = '\0';
+
+  if (name[0] == '\0' ||
+      strcmp (name, ".") == 0 ||
```
swick: I'm wondering if we can't allowlist things instead of catching bad cases. A bit uneasy about e.g. "foo/../../".

src/builder-utils.c
```diff
@@ -284,77 +284,244 @@ directory_is_empty (const char *path)
...
+                             GError **error)
+{
+  g_autoptr(GPtrArray) names = g_ptr_array_new_with_free_func (g_free);
 
-  while ((next = g_file_enumerator_next_file (dir_enum, NULL, &temp_error)))
+  {
+    g_auto(GLnxDirFdIterator) iter = { 0, };
+    struct dirent *dent;
+
+    if (!glnx_dirfd_iterator_init_at (src_dfd, ".", FALSE, &iter, error))
```
swick: Mh, `src_dfd` might come directly from `glnx_chase` which returns an `O_PATH` fd and at another place you re-open the directory which I think is correct.

src/builder-utils.c
```diff
@@ -284,77 +284,244 @@ directory_is_empty (const char *path)
...
+      struct glnx_statx stx;
+      char tmp_name[] = ".locale-XXXXXX";
+
+      glnx_gen_temp_name (tmp_name);
 
-      if (g_file_info_get_file_type (child_info) == G_FILE_TYPE_DIRECTORY)
+      if (!glnx_renameat (src_dfd, entry_name, src_dfd, tmp_name, error))
+        return FALSE;
+
+      chase_fd = glnx_chase_and_statxat (src_dfd, tmp_name,
```
swick: Same here: combine the loops, use chase first, rename then.

src/builder-utils.c
bbhtt: we need to skip it for C/en

src/builder-utils.c
bbhtt: I don't think a reopen is strictly needed here but i can add it.

src/builder-utils.c
bbhtt: same reason as below.

src/builder-utils.c
bbhtt: iirc this was about another race where multiple entries after the suffix is stripped could point to the same component.

src/builder-utils.c
bbhtt: no it cannot be allowlisted and the input here is d_name. I also realised now that glnx_dirfd_iterator_next_dent already filters this so it can be dropped.

src/builder-utils.c
bbhtt: it's split to collect first and mutate ie. rename later as i wasn't sure if iterating and renaming together would cause races or not.

src/builder-utils.c
swick: You want to avoid creating the directory if the locales are only C/en? I would suggest to filter out things in the first loop then.

bbhtt: > You want to avoid creating the directory if the locales are only C/en?

Yes

src/builder-utils.c
swick: The races exist no matter how you do it; code layout can never get rid of them, and inversely, there is no point in changing the code layout if you use proper fd based traversal.

src/builder-utils.c
swick: I'm fairly certain that readdir does not work on O_PATH fds because reading from a directory requires actual file permission checks.

src/builder-utils.c
bbhtt: I think it was about protecting my own operations while iterating and renaming not about external processes modifying the entries in-between.

src/builder-utils.c
bbhtt: I thought I was reopening at both call sites so I thought it isn't needed here but looks like I missed reopening the lang_dfd one.

src/builder-utils.c
bbhtt: skipped it in the first collection loop.

.github/workflows/check.yml
```diff
@@ -84,16 +84,6 @@ jobs:
       run: |
```
swick, line 84: commit message prefix missing

swick (approved):
Looks good now!

src/builder-utils.c
swick: Thanks for the review, let's not block on it though.

.github/workflows/check.yml
bbhtt: this is in a separate PR will land those and rebase.

bbhtt: https://github.com/flatpak/flatpak-builder/pull/760 is the PR, it's draft because I need to drop the wrap build steps first in https://github.com/flatpak/flatpak-builder/pull/752.

bbhtt:
I emailed @alexlarsson a while back about this but haven't heard back. It would be nice to get an extra look at this from him.

src/builder-utils.c
```diff
@@ -283,78 +283,228 @@ directory_is_empty (const char *path)
   return empty;
 }
 
+static void
+locale_name_to_language (char *name)
```
alexlarsson: This isn't currently wrong, but please make this take a const char *name, and return a char *, and them move the g_strdup() into this function.

bbhtt: Changed

src/builder-utils.c
```diff
@@ -283,78 +283,228 @@ directory_is_empty (const char *path)
...
+          return FALSE;
+        }
+
+      reopened_lang_dfd = glnx_fd_reopen (lang_dfd, O_RDONLY | O_DIRECTORY, error);
+      if (reopened_lang_dfd < 0)
+        return glnx_prefix_error (error, "failed to reopen %s", name);
+
+      glnx_gen_temp_name (tmp_name);
+
+      if (!glnx_renameat (source_dfd, name, source_dfd, tmp_name, error))
```
alexlarsson, line 500: I'm sligtlty worried about this, as there is no guarantees that the generated tmp name doesn't already exist as an empty dir which would make the renameat not error. However, I have not been able to come up with some case where this could end up causing problems. However, in some ideal world we would pass RENAME_NOREPLACE here.

alexlarsson:
lgtm

alexlarsson (approved):

src/builder-utils.c
bbhtt: > However, in some ideal world we would pass RENAME_NOREPLACE here.

there is glnx_renameat2_noreplace which we can use I guess but I too find it unlikely... also empty directories shouldn't cause problem here even if someone managed to place one.

bbhtt:
I have deployed this to Flathub just now, will try to backport based on the feedback from there in a couple of days.

PR #736, opened 2026-04-27
ci: Remove unused dependencies

bbhtt:
This cleans up old autotools dependencies and a bunch of dev packages which aren't necessary. libcurl-openssl is installed replacing gnutls as the former is the default used.


bbhtt:
Diffed the CI steps of configure and test between main and here, no relevant changes.

PR #737, opened 2026-05-13
ci: Exercise -Werror in CI

bbhtt:


bbhtt:
Depends on first commit in https://github.com/ostreedev/ostree/pull/3592 but I may have to apply that here too.

bbhtt:
#739, #740 and #741 is pre-req to enable this.

Issue #743, opened 2026-05-23
Setting sensible defaults for some CLI arguments


bbhtt:
- --state-dir - Default to a global state dir like `~/.cache/flatpak-builder` to enable sharing of downloaded artifacts and ccache artifacts across multiple manifests. This should speed up downloads and builds for someone building multiple things. The requirement is that the checkout app dir and state dir must stay on the same filesystem. This will also prevent people git add-ing `.flatpak-builder`.

- --ccache - Default it to on, every SDK ships ccache nowadays and if someone doesn't they should. See also https://github.com/flatpak/flatpak-builder/issues/582

- --subject - Set to SHA of head if inside a git directory. Already done by [flatpak-github-actions](https://github.com/flatpak/flatpak-github-actions/commit/ad1b66ed726a37baa034681da3009761063ec360) and [Flathub](https://github.com/flathub-infra/vorarbeiter/blob/932353cd0848f0a6fadb6e492f43ec7ebc457ec8/justfile#L18) since ages.

- --body - Set to something sensible like a hash of the manifest from state directory.

- --force-clean - Enable by default. Every new build should never reuse the checked out app dir and it currently errors anyway.

- --override-source-date-epoch - Set to a fixed date to enable reproducible builds by default. Manifest last modification timestamp is not reproducibile (see https://github.com/flathub/flathub/issues/2617) and a fixed date has been the standard for reproducible builds since a long time.

- --compose-url-policy - Default to 'full' to get back the appstream-glib behaviour which was lost during libappstream transition in 1.3.x. We will require and link against libappstream >=1.1.2 anyways.

- --user - Default to at least installing on the Flatpak user installation instead of system. Building and installing as user makes more sense and the system code path has been forever broken due to requiring auth etc https://github.com/flatpak/flatpak-builder/issues/290

For each of these options, the plan is to 

- change the default
- mark the old argument as deprecated (probably in the code and docs only and not any user facing warning)
- hide it from help
- introduce a negative `--disable-foo` where it makes sense
- update the documentation

Erick555:
> --force-clean - Enable by default. Every new build should never reuse the checked out app dir and it currently errors anyway.

Wouldn't it risk wiping content that wasn't supposed to be removed? I remember people wiping their entire home in the past but at least they explicitly passed this.

>  --user - Default to at least installing on the Flatpak user installation instead of system.

That would make it inconsistent with flatpak default behavior.

bbhtt:
[...]

PR #744, opened 2026-05-23
builder-main: Enable ccache by default if it exists in SDK

### LINKED ISSUE #582: make `--ccache` on by default
### Checklist

- [X] I have searched the [issue tracker](https://www.github.com/flatpak/flatpak-builder/issues) for a feature request that matches the one I want to file, without success.

### Suggestion

Passing `--ccache` is very helpful for development and it's annoying to ensure that it is passed to every invocation of `flatpak-builder`. It would be nice to have it on by default, or some other means of persistently enabling it (environment variable? AFAIK flatpak-builder has no persistent configuration file).

bbhtt:
The old `--ccache` argument is deprecated and replaced by a `--no-ccache` argument. It will be removed in some future unstable versions.

Now it is enabled by default if `ccache` exists in SDK or `--ccache` is explicitly passed while disabled if it does not exist or `--no-ccache` is explicitly passed.

In the first case, this also improves the previous behaviour and prevents blindly creating dangling symlinks when `ccache` does not actually exist in the SDK but someone passed `--ccache` anyway.

See: https://github.com/flatpak/flatpak-builder/issues/743
Fixes: https://github.com/flatpak/flatpak-builder/issues/582

### DIFF
--- a/src/builder-manifest.c
+++ b/src/builder-manifest.c
@@ -73,6 +73,7 @@ struct BuilderManifest
   char           *runtime_version;
   char           *sdk;
   char           *sdk_commit;
+  char           *sdk_path;
   char           *var;
   char           *base;
   char           *base_commit;
@@ -196,6 +197,7 @@ builder_manifest_finalize (GObject *object)
   g_free (self->runtime_version);
   g_free (self->sdk);
   g_free (self->sdk_commit);
+  g_free (self->sdk_path);
   g_free (self->base);
   g_free (self->base_commit);
   g_free (self->base_version);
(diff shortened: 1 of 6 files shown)


src/builder-manifest.c
```diff
@@ -1714,9 +1721,10 @@ builder_manifest_start (BuilderManifest *self,
              builder_manifest_get_runtime_version (self),
              self->sdk_commit);
 
-  sdk_path = flatpak_info_show_path (self->sdk, builder_manifest_get_runtime_version (self), context);
-  if (sdk_path != NULL &&
-      !builder_context_load_sdk_config (context, sdk_path, error))
+  g_free (self->sdk_path);
```
swick: g_clear_pointer

swick (approved):

src/builder-manifest.c
bbhtt: fixed

Be-ing:
Thanks for implementing my old feature request!

PR #752, opened 2026-05-28
build: Drop meson wraps

bbhtt:
- libglnx wrap is no longer needed since it became a subtree [1]
- libyaml is an optional dependency and widely available in distributions and we aren't actively looking for newer versions of it (development stopped for 5 years already but recently restarted). Moreover using it through a wrap raises a bunch of build warnings which are unfixed upstream. It is not exactly clear why it needed to be a wrap in [2]

[1]: https://github.com/flatpak/flatpak-builder/pull/692
[2]: https://github.com/flatpak/flatpak-builder/pull/382#discussion_r1007547265


bbhtt:
This and https://github.com/flatpak/flatpak-builder/pull/757 seems trivial to me, will merge soon.

swick (approved):

PR #753, opened 2026-05-28
builder-main: Set a default build subject and body

bbhtt:
See https://github.com/flatpak/flatpak-builder/issues/743

### DIFF
--- a/src/builder-utils.c
+++ b/src/builder-utils.c
@@ -2122,3 +2122,23 @@ appstream_has_version (int major,
          (as_major == major && as_minor > minor) ||
          (as_major == major && as_minor == minor && as_micro >= micro);
 }
+
+char *
+get_default_build_subject (void)
+{
+  g_autofree char *output = NULL;
+  const char *argv[] = { "git", "log", "-1",
+                         "--pretty=format:%s (%h)",
+                         "--abbrev=12", NULL };
+
+  if (!flatpak_spawnv (NULL, &output, G_SUBPROCESS_FLAGS_STDERR_SILENCE,
+                       NULL, argv, NULL))
+    return NULL;
+
+  if (output == NULL || output[0] == '\0')
+    return NULL;
+
+  /* Prevent non-ASCII characters from git commit subject in the
+     build subject */
+  return g_str_to_ascii (output, NULL);
+}
--- a/src/builder-utils.c
+++ b/src/builder-utils.c
@@ -2142,3 +2142,9 @@ get_default_build_subject (void)
      build subject */
   return g_str_to_ascii (output, NULL);
 }
+
+char *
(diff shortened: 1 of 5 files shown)


src/builder-utils.c
```diff
@@ -2122,3 +1973,27 @@ appstream_has_version (int major,
...
+
+  if (!flatpak_spawnv (NULL, &output, G_SUBPROCESS_FLAGS_STDERR_SILENCE,
+                       NULL, argv, NULL))
+    return NULL;
+
+  parts = g_strsplit (output, "\x1f", 2);
+  if (parts[0] == NULL || parts[1] == NULL)
+    return NULL;
+
+  g_autofree char *tmp = parts[0];
```
swick: why?

swick (commented):
Needs a proper rebase as well

src/builder-utils.c
bbhtt: something in the chain breaks if a non-ASCII character is passed as subject. 

we had this issue on flathub https://github.com/flathub/com.dec05eba.gpu_screen_recorder/issues/248

and i had to apply https://github.com/flathub-infra/vorarbeiter/blob/7d8a593b65a38646310e5cd18b1959b5e3f839f2/justfile#L24

I didn't have time to debug so it's a translation of what's being used.

bbhtt: the problematic character in that case was U+2026 ellipsis from the git subject.

bbhtt: Yea, I remember now it was in CI environment. Running it as 

```
LANG= flatpak-builder build --force-clean --ccache --user --subject="Open new ui if it's already started and trying to start it again. Sta… (#247)" test.yaml  
Option parsing failed: Invalid byte sequence in conversion input
```

fails locally.

swick: The entire function is just very confusion and needs a lot of explaining. `
  g_autofree char *tmp = parts[0];` also seems like a double free?

bbhtt: the idea is to now allow long subjects, I borrowed the git convention. the hash is 12 characters two braces and space so 15, and rest of the actual commit message is truncated to the remaining length.

smcv:
> something in the chain breaks if a non-ASCII character is passed as subject

Probably needs a UTF-8 locale and a call to `setlocale()`?

bbhtt:
Maybe, or just simply setting LANG= specifically works. All LC_*` variables was already set to UTF-8 but seems LANG is needed.

src/builder-utils.c
bbhtt: dropped the truncation, it might have been a bit of over-engineer, but we don't do it in the flathub case anyways and it indeed looks a bit confusing..

swick (approved):

PR #755, opened 2026-05-30
tests: Add tests for locale migration

bbhtt:


swick (approved):
I don't know the details of locale migration but the tests seem to exercise the current behavior, so LGTM

bbhtt:
Using this as a baseline for fd rewrite so the main thing I am checking is to not miss an usecase. There's this structure and two more tests in test-builder / test-locale-cleanup where files are flat. These are the gettext convention.

PR #765, opened 2026-07-18
builder-context: Chain-up constructed vfunc

A6GibKm:
Via

    gobject-linter --only g_object_virtual_methods_chain_up --fix


TingPing:
It's fine but I'd expect the chain up to be at the start or the end

A6GibKm:
Done. Yeah, it was after the definitions, but it looked odd.

PR #767, opened 2026-08-08
builder: Add AppStream compose icon policy option

razzeee:
Hopefully allows us to start generating bigger icons on flathub if we want that


bbhtt:
1. Wrong commit prefix and no commit message
2. Variable names and arguments not following the pattern in use
3. There is no need to checksum cache
4. The way it is implemented this is not backwards compatible 
5. AI slop comments in tests The first time I hear of "retina" and "non-retina" icons. 

In general if you are unable to distil and add human moderation to the AI output I cannot be bothered. You can just open an issue and wait. This also needs to wait on https://github.com/flatpak/flatpak-builder/pull/664 anyways.

razzeee:
I'm sorry I called it retina, I'm blanking on the naming. That's what I usually use when talking to designers, as they know it from macOS.

What about the names is not correct? I changed them to start with `opt_` specifically cause there is old code like that.

razzeee:
Updated accordingly as far as I got your comment, kinda hard to get meaning out of it in most places, let me know if it moved into the correct direction.

PR #768, opened 2026-08-22
Drop debugedit submodule

A6GibKm:
Meson uses system debugedit.

Backported from main.


A6GibKm:
We are having issues at https://github.com/flathub/org.gnome.Builder/pull/91 when downloading this module. 

I am not sure if this is a good idea though.

bbhtt:
1.4.x still supports autotools. It's the reason why this wasn't backported.

bbhtt:
See https://github.com/flathub/org.flatpak.Builder/blob/7f914ae61491874564d45ccaf11c26c675447b5b/org.flatpak.Builder.json#L108 for a workaround. libglnx is rarely bumped so pinning it independently won't be an issue and if we start depending on new API the build will fail anyways.

A6GibKm:
Thanks a lot for the hint. I am closing this.

PR #773, opened 2026-09-06
builder-utils: Skip pre-migrated entry during locale migration

### LINKED ISSUE #772: Invalid cross-device link
1.4.10 seems to have introduced a regresssion.

### Flatpak builder version
1.4.10

### Flatpak version
1.18.0

### Full commandline arguments used
`flatpak-builder --verbose --disable-updates --disable-rofiles-fuse --ccache --user --install-deps-from=flathub build3 org.kde.Sdk.json.in`

### Permanent link to all manifest and build sources (if applicable)
https://invent.kde.org/packaging/flatpak-kde-runtime/-/blob/work/justin/qt610-updates-2026-09-01/org.kde.Sdk.json.in?ref_type=heads

### Description of the issue
Invalid cross-device link

### Logs
Invalid cross-device link

1.4.9 works fine, 1.4.10 does not. Maybe this https://github.com/flatpak/flatpak-builder/commit/01d55c7e01520340433fde9677735de318baf495

Edit: Wrong link inserted.

bbhtt:
builder-utils: Skip pre-migrated entry during locale migration

A parent SDK can ship locale content that is already split out into
`share/runtime/locale` structure leaving a relative symlink
`../../share/runtime/locale/<lang>/...` behind. Since the locale
content of a child SDK is composed of locale content of the parent SDK
and its own, we chase both types of entries with chaseat's
RESOLVE_BENEATH.

This is incorrect when we encounter a pre-migrated entry belonging
to parent SDK as it resolves outside the root already. chaseat with
RESOLVE_BENEATH raises EXDEV here and trips up the entire migration.
So let's skip such pre-migrated entries.

An entry is considered "pre-migrated" if after reading the symlink,
it string matches the expected layout ie. when iterating
`root_dir/prefix/share/locale`, symlinks to `../runtime/locale` are
skipped and when iterating `root_dir/prefix/lib/locale` symlinks
to `../../share/runtime/locale` are skipped. FDSDK ships with both
kinds of entries.

Fixes: https://github.com/flatpak/flatpak-builder/issues/772


src/builder-utils.c
```diff
@@ -451,6 +451,26 @@ migrate_locale_dir (int         root_dfd,
...
+        return glnx_throw_errno_prefix (error, "fstatat %s", dent->d_name);
+
+      if (S_ISLNK (st.st_mode))
+        {
+          g_autofree char *language = locale_name_to_language (dent->d_name);
+          const char *expected_target_prefix = strcmp (source_rel, "share/locale") == 0 ?
+            "../runtime/locale" :
+            "../../share/runtime/locale";
+
+          g_autofree char *expected_target = g_build_filename (expected_target_prefix,
```
swick: Not a fan of the complex logic in variable declarations

src/builder-utils.c
```diff
@@ -451,6 +451,26 @@ migrate_locale_dir (int         root_dfd,
       if (dent == NULL)
         break;
 
+      struct stat st;
```
swick: declarations at the top of a scope

src/builder-utils.c
```diff
@@ -451,6 +451,26 @@ migrate_locale_dir (int         root_dfd,
...
+
+      if (fstatat (source_dfd, dent->d_name, &st, AT_SYMLINK_NOFOLLOW) < 0)
```
swick: glnx provides wrappers for those handling edge cases and the error.

swick (commented):
Makes sense, code could use some cleanup.

src/builder-utils.c
```diff
@@ -451,6 +451,26 @@ migrate_locale_dir (int         root_dfd,
...
+          g_autofree char *language = locale_name_to_language (dent->d_name);
+          const char *expected_target_prefix = strcmp (source_rel, "share/locale") == 0 ?
+            "../runtime/locale" :
+            "../../share/runtime/locale";
+
+          g_autofree char *expected_target = g_build_filename (expected_target_prefix,
+                                                               language, subdir, dent->d_name, NULL);
+          g_autofree char *actual_target = glnx_readlinkat_malloc (source_dfd, dent->d_name, NULL, error);
+
+          if (actual_target && strcmp (actual_target, expected_target) == 0)
```
swick: `g_strcmp0` instead of `strcmp` everywhere

src/builder-utils.c
bbhtt: split the initialisation.

src/builder-utils.c
bbhtt: moved at top

src/builder-utils.c
bbhtt: switched to glnx_fstatat

src/builder-utils.c
```diff
@@ -419,6 +419,7 @@ migrate_locale_dir (int         root_dfd,
   glnx_autofd int separate_dfd = -1;
   g_auto(GLnxDirFdIterator) iter = { 0, };
   struct dirent *dent;
+  struct stat st;
```
swick, line 422: can move to the smaller scope

src/builder-utils.c
```diff
@@ -445,20 +446,45 @@ migrate_locale_dir (int         root_dfd,
...
+
+      language = locale_name_to_language (dent->d_name);
+
+      if (S_ISLNK (st.st_mode))
+        {
+          const char *expected_target_prefix;
+          g_autofree char *expected_target = NULL;
+          g_autofree char *actual_target = NULL;
+
+          if (strcmp (source_rel, "share/locale") == 0)
```
swick, line 468: `g_strcmp0`

swick (approved):
Some suggestions, but feel free to merge

bbhtt:
Thanks fixed the final two nits.

Issue #776, opened 2026-09-14
Tests fail with libappstream 1.2.0


smcv:
### Flatpak builder version

1.4.10, 1.4.11, git main at 58c03948955b4df8cf7b61e1e984a2b50688f4e9

### Flatpak version

1.8.2

### Full commandline arguments used

`meson test -C _build`

### Permanent link to all manifest and build sources (if applicable)

See test suite in source tree

### Description of the issue

Since Debian updated from libappstream 1.1.6 to 1.2.0, if the flatpak-builder test suite is run in an environment where `bwrap` is runnable (e.g. real hardware or virtual machine, but not the containers used by the official Debian autobuilders), it fails. This can happen at build-time, or when the as-installed tests are run via gnome-desktop-testing-runner.

Specifically, the part of `test-builder.sh` after outputting `ok screenshot ref commit` and before `ok compose partial url policy` fails. The part from `ok compose partial url policy` to `ok compose full url policy` also fails.

### Logs

The log doesn't show anything particularly comprehensible - perhaps it's the `grep` that is failing (with no output).

```
...
123s ok screenshot ref commit
123s Emptying app dir 'builddir_sc'
123s Downloading sources
123s Initializing build dir
123s Committing stage init to cache
123s Starting build of org.flatpak.appstream_media
123s ========================================================================
123s Building module appstream_media in /var/tmp/test-flatpak-0gQYxC/.fp-compose-url-policy-partial/build/appstream_media-1
123s ========================================================================
123s Running: mkdir -p ${FLATPAK_DEST}/bin ${FLATPAK_DEST}/share/metainfo ${FLATPAK_DEST}/share/applications
123s Running: mkdir -p ${FLATPAK_DEST}/share/icons/hicolor/256x256/apps
123s Running: cp -vf hello.sh ${FLATPAK_DEST}/bin/hello
123s 'hello.sh' -> '/app/bin/hello'
123s Running: cp -vf org.flatpak_builder.gui.metainfo.xml ${FLATPAK_DEST}/share/metainfo
123s 'org.flatpak_builder.gui.metainfo.xml' -> '/app/share/metainfo/org.flatpak_builder.gui.metainfo.xml'
123s Running: cp -vf org.flatpak_builder.gui.desktop ${FLATPAK_DEST}/share/applications
123s 'org.flatpak_builder.gui.desktop' -> '/app/share/applications/org.flatpak_builder.gui.desktop'
[...]

PR #778, opened 2026-09-14
tests: Use "no plan" style of TAP output

smcv:
* tests: Use "no plan" style of TAP output
    
    Similar to
    https://github.com/flatpak/flatpak/commit/b1fdf6c4ef9d10b90a5d1004963f9c13606e2c15
    this makes it more obvious where the boundary between test-cases is,
    and similar to https://github.com/flatpak/flatpak/pull/6794 it avoids
    needing to declare up-front how many tests we plan to run (which makes
    conflicts inevitable if we add more than one test to the same script).


tests/test-sandboxed-mode-source-resolution.sh
```diff
@@ -481,5 +479,6 @@ BUILD_LOG=build-log-git-file_uri-outside run_build_fail --sandbox \
     test-git-file_uri-outside.json
 assert_file_has_content build-log-git-file_uri-outside 'not inside manifest directory'
 
-echo "ok file URI source outside manifest tree is rejected for git type in sandboxed mode"
+ok "file URI source outside manifest tree is rejected for git type in sandboxed mode"
 
+done_testing
```
bbhtt, line 484: is this done_testing intentional or left over?

bbhtt:
LGTM minus the JXL change in appstream as I said in the other thread. 

If you want this merged first, you can drop the JXL change. 

I'll have to do some changes to the code to make sure it stays PNG and release a new version.

tests/test-sandboxed-mode-source-resolution.sh
bbhtt: ah nevermind i missed the function.

smcv:
I moved the most important part of this for diagnosing #776 into #781.

Because the number of tests in `test-builder.sh` has changed between `flatpak-builder-1.4.x` and `main`, it's not going to be useful to rebase `tests: Use "no plan" style of TAP output` onto `main` until the two branches have been reconciled.

bbhtt:
Are you going to rebase this? or should I take over? It's the last thing before the 1.4.12 release.

smcv:
This will need a rebase after #782 lands, but is otherwise ready (only look at the last commit).

This also doesn't have to be a release blocker, skipping it until after 1.4.12 is completely fine.

I've targeted `main` only here: I don't know whether you'll want it in 1.4.x or not. If cherry-picked to 1.4.x, there will be some conflicts in places where `main` has added additional test coverage (just `test-builder.sh`, I think).

smcv:
eaa42a18: rebased after #782 landed, no functional change.

bbhtt:
Yep I am backporting it to 1.4.x

PR #777, opened 2026-09-14
test-builder.sh: Allow Appstream icons to be in PNG or JPEG-XL format

smcv:
Older versions of the Appstream library always used PNG here, but 1.2.0 uses JPEG-XL.

Resolves: https://github.com/flatpak/flatpak-builder/issues/776

PR #779, opened 2026-09-15
builder-main: Allow specifying appstream-compose image format

bbhtt:
This additionally ensures the default format stays PNG consistent with
pre-1.2.0 behaviour and to preserve backwards compat while allowing
JXL if the required AppStream version (>=1.2.0) is available.

Related: https://github.com/flatpak/flatpak-builder/pull/777
Related: https://github.com/flatpak/flatpak-builder/issues/776
Closes: https://github.com/flatpak/flatpak-builder/issues/776
Closes: https://github.com/flatpak/flatpak-builder/pull/777

Can't find CI with 1.2.0 yet so haven't added tests...


bbhtt:
Yeah will have to add 1.2.0 CI later on. Building 1.2.0 in CI is complicated. It needs a newer versions of a bunch of dependencies gcc meson and some new dependencies too which aren't in the 22.04/24.04 repos yet.

I'll probably switch to newer Ubuntu or Fedora once they have it.

bbhtt:
I added a tentative test for JXL in the last commit. Needs to be verified it is working.

smcv:
> I'll probably switch to newer Ubuntu or Fedora once they have it

Ubuntu LTS is unlikely to pick up a new `appstream` package in anything less than a new major release (the next one will be 28.04 in April 2028). To test against the latest software, you'll have to use a development or rolling-release distribution like Debian unstable, Fedora rawhide or Arch, or at least a short-term-support release like Fedora.

smcv (changes requested):
The changes generally good to me, but the tests are failing on Debian unstable. I'll see if I can get more useful debug info out of them.

Are you intending to merge this into main as well as 1.4.x, when it works? (I ask so that I can try to rebase #778 to get useful diagnosis.)

bbhtt:
Yes I do intend to merge on both branches. If you can get the info on why the tests are failing it would be great.

> you'll have to use a development or rolling-release distribution like Debian unstable, 

Yep, but only Debian unstable has it right now and it is a bit of a pain to use in CI as it's not a full repository. The unstable distros/branches are also a bit too rolling for CI to my taste.

I intend on switching to Fedora when they'll have it.

smcv:
> not a full repository

Er, it is? Debian unstable is a complete, self-contained distribution just like stable.

Perhaps you're thinking of experimental, which is an overlay over unstable, the same way that stable-backports is an overlay over stable?

bbhtt:
Ah right I confused it with experimental

smcv:
> If you can get the info on why the tests are failing it would be great

Half of it was a timeout, fixed (or at least worked around) by #780. I'm still struggling with the shell scripts to figure out why `test-appstream.sh` is failing.

bbhtt:
Ok thanks. I'll try to figure it out if you don't have the time.

tests/test-appstream.sh
```diff
@@ -0,0 +1,83 @@
...
+# test compose jxl image format
+if appstream_has_version 1 2 0; then
+    APPDIR=builddir_jxl \
+    run_build \
+        --mirror-screenshots-url=https://example.org/media \
+        --state-dir .fp-compose-image-format-jxl \
+        --compose-image-format=jxl \
+        org.flatpak.appstream_media.json
+
+        gzip -cdq builddir_jxl/files/share/app-info/xmls/org.flatpak.appstream_media.xml.gz|grep -Eq '>https://example.org/media/org/flatpak/appstream_media/[^/]+/icons/128x128/org.flatpak.appstream_media.jxl</icon>'
```
smcv, line 79: With `assert_file_has_content_extended` patched in from #778:

```
# <?xml version="1.0" encoding="utf-8"?>                                          
# <components version="1.2" media_baseurl="https://example.org/media">            
#   <component type="desktop-application">                                        
#     <id>org.flatpak.appstream_media</id>                                        
#     <name>org.flatpak_builder.gui</name>                                        
#     <summary>Foo foo foo foo</summary>                                          
#     <project_license>GPL-2.0+</project_license>                                 
#     <description>                                                               
#       <p>An example desktop application</p>                                     
#     </description>                                                              
#     <developer>                                                                 
#       <name>Flatpak</name>                                                      
#     </developer>                                                                
#     <launchable type="desktop-id">org.flatpak.appstream_media.desktop</launchable>
#     <icon type="cached" width="64" height="64">org.flatpak.appstream_media.jxl</icon>
#     <icon type="remote" width="128" height="128">org/flatpak/appstream_media/85dc9b56ceb35841b789bd23034d1b98/icons/128x128/org.flatpak.appstream_media.jxl</icon>
#     <icon type="cached" width="128" height="128">org.flatpak.appstream_media.jxl</icon>
#     <icon type="stock">org.flatpak.appstream_media</icon>                       
#     <url type="homepage">https://flatpak.org</url>                              
#     <url type="bugtracker">https://flatpak.org</url>                            
#     <categories>                                                                
#       <category>Network</category>                                              
#     </categories>                                                               
#     <provides>                                                                  
#       <id>org.flatpak_builder.gui</id>                                          
#     </provides>                                                                 
#     <releases>                                                                  
#       <release type="stable" version="0.0.1" timestamp="1598572800"/>           
#     </releases>                                                                 
#     <content_rating type="oars-1.1"/>                                           
#   </component>                                                                  
# </components>                                                                   
File 'appstream.xml' doesn't match extended regexp '>https://example.org/media/org/flatpak/appstream_media/[^/]+/icons/128x128/org.flatpak.appstream_media.jxl</icon>'
```

I think you need to either pass `--compose-url-policy=full` to this third `run_build` invocation, or change the regex so it expects `>org/flatpak/…` like the one for the first `run_build` invocation.

bbhtt: ah yep that's it.

smcv (approved):
Now works (when rebased on #780), thanks!

bbhtt:
Ok thanks. I'll merge this and merge your test improvement PR then cut a release. 



I'll open an issue to track the missing CI part.

PR #782, opened 2026-09-15
Cherry-pick #779, #781 to main

bbhtt:


smcv (commented):
1.4.x to 1.5.x is surely a forward-port rather than a backport? But the code changes look about right. Let me try this on a system that has appstream 1.2.0.

smcv (approved):
Yes, seems fine.

I think it would make sense to cherry-pick 72b723402a76ba6b44dcbdbbd8ec9cf03cb5c4d9 and 1da33b6e4156a2571cc2d1e44e658863681b6d35 (#781) as part of this. They cherry-pick cleanly, and the tests pass on appstream 1.2.0 with or without them.

bbhtt:
Thanks added those commits.

smcv (approved):

PR #784, opened 2026-09-15
[1.4] tests: Use "no plan" style of TAP output

bbhtt:
Backport #778


tests/test-builder-licence-paths.sh
```diff
@@ -153,4 +151,4 @@ assert_file_has_content \
     appdir/files/share/licenses/org.test.licence_working_2/test/LICENSE \
     '^MY DEFAULT LICENSE TEXT$'
 
-echo "ok license file is recorded by default"
+done_testing
```
smcv, line 156: I would have expected this to turn into

```suggestion
ok "license file is recorded by default"

done_testing
```

bbhtt: thanks, looks like i removed it while fixing the conflict.

smcv (approved):

PR #786, opened 2026-09-15
[1.4.x] builder-module: Abort on invalid module names

bbhtt:
Backports: https://github.com/flatpak/flatpak-builder/pull/715


swick (approved):

bbhtt:
do we need a report for the formalities?

bbhtt:
> do we need a report for the formalities?

ping @swick

swick:
Probably not a bad idea, I think this might already have a CVE assigned even?

bbhtt:
I'm not sure which is it, I don't have the report I think.

swick:
Never mind, only a RH internal ticket exists for this.

smcv:
I don't see anything in https://github.com/flatpak/flatpak-builder/security/advisories, which ideally should list everything, or https://security-tracker.debian.org/tracker/source-package/flatpak-builder, which is usually fairly comprehensive for CVEs that are known to the public.

@swick, would you be able to get a CVE ID from Red Hat referencing the internal ticket? Probably best if you do it, via RH as CNA, because otherwise they'll allocate a duplicate CVE ID for their internal ticket and everyone will become confused.

swick:
I can, but it would be good to be able to just reference a published advisory.

